PR #1, opened 2018-06-15
"any" is in List::Util and was added in version 1.33

muffato:


james-monkeyshines:
List::Util is a core module, I thought; what version of Perl are you using?

muffato:
Ah, I didn't know ! I'm on Perl 5.18 and it's missing `any`.

Is `ensembl-datacheck` supposed to be compatible with Perl 5.14 as per @nerdstrike  's email ?

james-monkeyshines:
`any` is in List::Util for 5.14. But it's part of a fundamental portion of code, so ensuring it is there whatever your perl version seems sensible.

PR #11, opened 2018-08-30
New datacheck to check that the version in the meta table is consistent with the database name

james-monkeyshines:


james-monkeyshines:
Picked up a couple of changes in index.json here that relate to another datacheck. Fine for now, but needs to be addressed. It happens because the datacheck meta data was edited after datacheck creation; create_index.pl needs to be re-run in such cases, but that's not written in the guidelines. Should also have a test for that.

PR #12, opened 2018-08-31
New tests/check char

marcoooo:
Added CheckChar to DC. Please carefully review, as this is my first Perl script in years!

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
@@ -24,6 +24,8 @@ use strict;
 use Moose;
 use Test::More;
 
+use 'Bio::EnsEMBL::DataCheck::Test::DataCheck';
+
 extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 
 use constant {
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
@@ -23,8 +23,7 @@ use strict;
 
 use Moose;
 use Test::More;
-
-use 'Bio::EnsEMBL::DataCheck::Test::DataCheck';
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
 
 extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
@@ -1,0 +1,83 @@
+=head1 LICENSE
+
+Copyright [2018] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckChar;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+    NAME        => 'CheckChar',
+    DESCRIPTION => 'Check that imported names/descriptions contains only supported characters',
+};
+
+sub tests {
+    my ($self) = @_;
+
+    # Test description length
+    my $sql_length = qq/
+        SELECT name
+        FROM phenotype
+        WHERE description IS NOT NULL
+        AND LENGTH(description) < 4
+    /;
+    my $desc_length = 'Phenotype description length';
+    my $diag_length = "Phenotype is suspiciously short";
+    is_rows_zero($self->dba, $sql_length, $desc_length, $diag_length);
+
+    # Test description new line character
+    my $sql_newline = qq/
+        SELECT name
+        FROM phenotype
+        WHERE description IS NOT NULL
+        AND description LIKE '%\n%'
+    /;
+    my $desc_newline = 'Phenotype description with new line';
+    my $diag_newline = "Phenotype is suspiciously short";
+    is_rows_zero($self->dba, $sql_newline, $desc_newline, $diag_newline);
+
+    # Test ASCII vars
+    my $sql_ascii = qq/
+        SELECT name, description
+        FROM phenotype
+        WHERE description NOT REGEXP '[ -~]'
+        OR description REGEXP '.*\<.*\>.*'
+    /;
+    my $desc_ascii = 'Phenotype description with unsupported character';
+    my $diag_ascii = "Phenotype has suspect start or unsupported characters";
+    is_rows_zero($self->dba, $sql_ascii, $desc_ascii, $diag_ascii);
+
+    # Check non terms in description
+    my $sql_non_term = qq/
+        SELECT name, description
+        FROM phenotype
+        WHERE lower(description) in ("none", "not provided", "not specified", "not in omim", "variant of unknown significance", "not_provided", "?", ".")
+    /;
+    my $desc_non_term = 'Phenotype description suggests no phenotype';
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+    NAME        => 'CheckChar',
+    DESCRIPTION => 'Check that imported names/descriptions contains only supported characters',
```
james-monkeyshines: This is only relevant for variation databases, so need DB_TYPES => ['variation'] here.
Also, since this datacheck is only concerned with one table, that should be set (i.e. TABLES => ['phenotype']) in order for efficient skipping of tests.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+};
+
+sub tests {
+    my ($self) = @_;
+
+    # Test description length
+    my $sql_length = qq/
+        SELECT *
```
james-monkeyshines: In the case of failure, all SELECTED columns will be returned as a diagnostic message, comma-separated; so "SELECT *" is fine if that is what's intended. There are only 4 columns in this table, so that's not unreasonable. I would be inclined to do "SELECT description", because then it's clear what was column was being tested.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+        FROM phenotype
+        WHERE description IS NOT NULL
+        AND LENGTH(description) < 4
+    /;
+    my $desc_length = 'Phenotype description length';
+    my $diag_length = "Phenotype is suspiciously short";
```
james-monkeyshines: I think "Description is suspiciously short" would be a bit clearer.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+
+    # Test description new line character
+    my $sql_newline = qq/
+        SELECT *
+        FROM phenotype
+        WHERE description IS NOT NULL
+        AND description LIKE '%\n%'
+    /;
+    my $desc_newline = 'Phenotype description with new line';
+    my $diag_newline = "Phenotype is suspiciously short";
```
james-monkeyshines: Diag text copied and pasted from above...

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+    my ($self) = @_;
+
+    # Test description length
+    my $sql_length = qq/
+        SELECT *
+        FROM phenotype
+        WHERE description IS NOT NULL
+        AND LENGTH(description) < 4
+    /;
+    my $desc_length = 'Phenotype description length';
```
james-monkeyshines: Description (and diag) variable should go above the SQL variable. Then you do not need the comment on line 38. (Comment also applies to other tests.)

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+    is_rows_zero($self->dba, $sql_length, $desc_length, $diag_length);
+
+    # Test description new line character
+    my $sql_newline = qq/
+        SELECT *
+        FROM phenotype
+        WHERE description IS NOT NULL
+        AND description LIKE '%\n%'
+    /;
+    my $desc_newline = 'Phenotype description with new line';
```
james-monkeyshines: This is not what we are checking (the logic needs to be flipped from what is in the HC); we need something that is true for passing tests, e.g. "No newlines in phenotype description"

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckChar;
```
james-monkeyshines: In general, I think it's good to have datachecks with the same name as healthchecks, to help work out equivalencies. But here I'm not sure if that's outweighed by the usefulness of changing to a more meaningful name, e.g. PhenotypeDescriptions; CheckChar is both inaccurate and very generic.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+
+    # Test ASCII vars
+    my $sql_ascii = qq/
+        SELECT *
+        FROM phenotype
+        WHERE description REGEXP '[^ -;=\?-~]'
+        OR LEFT(description, 1) REGEXP '[^A-Za-z0-9]'
+
+    /;
+    my $desc_ascii = 'Phenotype description with unsupported character';
```
james-monkeyshines: Flip logic, as above, suggest "No unsupported or suspicious characters in phenotype description"

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+    my $diag_ascii = "Phenotype has suspect start or unsupported characters";
+    is_rows_zero($self->dba, $sql_ascii, $desc_ascii, $diag_ascii);
+
+    # Check non terms in description
+    my $sql_non_term = qq/
+        SELECT *
+        FROM phenotype
+        WHERE lower(description) in ("none", "not provided", "not specified", "not in omim", "variant of unknown significance", "not_provided", "?", ".")
+    /;
+    my $desc_non_term = 'Phenotype description suggests no phenotype';
```
james-monkeyshines: Flip logic, as above, suggest "Meaningful phenotype description"

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
```diff
@@ -0,0 +1,85 @@
...
+    # Test ASCII vars
+    my $sql_ascii = qq/
+        SELECT *
+        FROM phenotype
+        WHERE description REGEXP '[^ -;=\?-~]'
+        OR LEFT(description, 1) REGEXP '[^A-Za-z0-9]'
+
+    /;
+    my $desc_ascii = 'Phenotype description with unsupported character';
+    my $diag_ascii = "Phenotype has suspect start or unsupported characters";
```
james-monkeyshines: Diag text does not need to be so wordy; we already know something is wrong from the description that will be printed immediately above. This is more about describing what data is being returned as an example of the violation. If you stick with "SELECT *" and return all columns, then something like "phenotype row" would make more sense; for a "SELECT description" statement, you could use simply use "Description".

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
marcoooo: Well, there is effectively only 4 column, I think only description could not help finding data from report in DB. Le't keep it that way

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
marcoooo: This was the message issued from original HC, just added length information.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
marcoooo: Done

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
marcoooo: Well, no worries. Changed name to PehnotypeDescriptions

lib/Bio/EnsEMBL/DataCheck/Checks/CheckChar.pm
marcoooo: Fine

PR #13, opened 2018-08-31
Skip foreign key check in case of polymorphic associations

MatBarba:
This happens when several foreign keys are defined from the same field.

This is currently the case in the variation phenotype_feature table, field object_id. Though ideally the schema should be changed to avoid those.


james-monkeyshines:
Thanks for pointing this out. I don't think it makes sense to have such foreign keys defined at all. If you created innodb tables with these constraints and tried to populate the tables it wouldn't work. So in the first instance I'll ask variation team to remove them. Dependent constraints such as these can be checked separately, a la the CoreForeignKeys datacheck, where object_xref has a similar situation.

james-monkeyshines:
Fixed at source by the variation team: https://github.com/Ensembl/ensembl-variation/commit/4234521

PR #14, opened 2018-09-03
New datacheck to make sure that every epigenome which is part of the …

thomasmaurel:
…current Regulatory Build has a segmentation file in the segmentation_file table.
Port of the following healthcheck: https://github.com/Ensembl/ensj-healthcheck/blob/release/94/src/org/ensembl/healthcheck/testcase/funcgen/EpigenomeHasSegmentationFile.java
I wrote two ways, one using the API and one using SQL. I think the API one is nicer but it's slightly slower by 1-2s. We might not want to keep both, it's your call ;-)

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
@@ -31,7 +31,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME        => 'EpigenomeHasSegmentationFile',
   DESCRIPTION => 'Check that every epigenome which is part of the current Regulatory Build has a segmentation file in the segmentation_file table',
-  GROUPS      => ['funcgen_integrity', 'funcgen_Post_regulatory_build.'],
+  GROUPS      => ['funcgen_integrity', 'funcgen_Post_regulatory_build'],
   DB_TYPES    => ['funcgen'],
   TABLES      => ['regulatory_build','regulatory_build_epigenome','epigenome'],
 };
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
@@ -1,0 +1,103 @@
+=head1 LICENSE
+
+Copyright [2018] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::EpigenomeHasSegmentationFile;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use Bio::EnsEMBL::DataCheck::Utils qw/sql_count/;
+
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
```diff
@@ -0,0 +1,103 @@
...
+  DB_TYPES    => ['funcgen'],
+  TABLES      => ['regulatory_build','regulatory_build_epigenome','epigenome'],
+};
+
+sub skip_tests {
+  my ($self) = @_;
+
+  my $sql = q/
+    SELECT COUNT(name) FROM regulatory_build 
+    WHERE is_current=true
```
james-monkeyshines: is_current is a tinyint column, so should be 'WHERE is_current = 1'

lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
```diff
@@ -0,0 +1,103 @@
...
+  /;
+
+  if (! sql_count($self->dba, $sql) ) {
+    return (1, 'The database has no regulatory build');
+  }
+}
+
+sub tests {
```
james-monkeyshines: I would have written the tests method quite differently...

Firstly, while using the API is generally to be preferred, if, as in this case, the API doesn't have all the requisite functionality, and you end up having to do SQL queries anyway, then I'd be inclined to go with exclusively SQL-based code, rather than a bit of both. Conceptually it seems a bit cleaner.

Secondly, I understand why you use a for loop with a $pass variable, because this avoids doing many tests, but it seems somewhat inelegant. Also, you're doing a query per table row, which isn't too bad in this case, because there's not much data, but ideally we'd want to minimise the number of queries. The is_rows_zero method is almost doing what we want; it does one query, and reports any rows that are erroneously returned as diagnostic messages. The task, then, is to construct that one query, such that we don't need to loop over epigenomes; while we're at it, we can eliminate the need to fetch the regulatory_build_id, and just use the current one.

The content of the tests method could then be more succinct:
my ($self) = @_;
my $desc = "All the epigenomes of the current regulatory build have a segmentation file";
my $diag = "Segmentation file missing for epigenome_id";
my $sql = q/
  SELECT
    e.epigenome_id
  FROM
    regulatory_build rb INNER JOIN
    regulatory_build_epigenome rbe ON rb.regulatory_build_id = rbe.regulatory_build_id INNER JOIN
    epigenome e ON rbe.epigenome_id = e.epigenome_id LEFT OUTER JOIN
    segmentation_file sf ON e.epigenome_id = sf.epigenome_id
  WHERE
    rb.is_current = 1 AND
    sf.epigenome_id is null
/;
is_rows_zero($self->dba, $sql, $desc, $diag);

This single query is more complicated, but my opinion (and please feel free to disagree...) is that it's easier to see what's going on, because it's all in once place, not spread over three queries. Outer joins aren't always the quickest way to do something, but there's not much data here so it's not an issue. An alternative would be to extract two lists of epigenome_ids (one from the rb/rbe/e join, one from sf) and compare them with an is_deeply test; that method will automatically give you a report of discrepancies between the lists.

Please let me know if any of this seems daft/counterintuitive/wrong...

lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
thomasmaurel: Well spotted. In my next commit, I will also update CurrentRegulatoryBuildHasEpigenomes.

lib/Bio/EnsEMBL/DataCheck/Checks/EpigenomeHasSegmentationFile.pm
thomasmaurel: Thanks a lot for the detailed explanation. This makes perfect sense to me and is much cleaner. Indeed I uses a pass variable to avoid getting a report for each row. I had a bit of trouble doing this one as I couldn't find a similar example but I didn't think of merging everything into one SQL. 
I am actually doing something similar in my next test by using this SQL directly instead of looping through object: https://github.com/Ensembl/ensj-healthcheck/blob/release/94/src/org/ensembl/healthcheck/testcase/funcgen/RegulatoryFeatureIsActive.java#L90. So things will be more consistent like this.
Thanks a again!!!

thomasmaurel:
@james-monkeyshines: Pushed all requested changes.

PR #16, opened 2018-09-05
New test/regulatory feature is active

thomasmaurel:
Port of the following healthcheck: https://github.com/Ensembl/ensj-healthcheck/blob/release/94/src/org/ensembl/healthcheck/testcase/funcgen/RegulatoryFeatureIsActive.java
I used the final SQL instead of looping through the different objects. The hc mention that the SQL is slow but it only takes 3s for human. Ilias is having a look why he didn't used the final SQL in the hc, if he comes back to me, I will update the test.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/index.json
+++ b/lib/Bio/EnsEMBL/DataCheck/index.json
@@ -102,7 +102,7 @@
       "description" : "Check if the current regulatory build has epigenomes data",
       "groups" : [
          "funcgen_integrity",
-         "funcgen_Post_regulatory_build."
+         "funcgen_Post_regulatory_build"
       ],
       "name" : "CurrentRegulatoryBuildHasEpigenomes",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::CurrentRegulatoryBuildHasEpigenomes"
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -191,6 +191,13 @@
       "name" : "PhenotypeDescription",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::PhenotypeDescription"
    },
+   "RegulatoryFeatureIsActive" : {
+      "datacheck_type" : "critical",
+      "description" : "Check that every regulatory_feature of the current Regulatory Build has a valid activity value in at least one epigenome.",
+      "groups" : [],
```
james-monkeyshines: The groups in the index don't match those in the datacheck (which is why travis is failing). Presumably you added the groups after you had run the creation script. I belatedly added a note to the docs that in such cases you need to run create_index.pl (https://www.ebi.ac.uk/seqdb/confluence/display/GTI/Everything+you+wanted+to+know+about+datachecks%2C+but+were+too+afraid+to+ask#Everythingyouwantedtoknowaboutdatachecks,butweretooafraidtoask-Creatingadatachecktemplate)

But it's quite annoying to have to remember to do this, when it seems like it could be automated. I've never used git hooks, but I'll look into whether they could be useful here.

thomasmaurel: Arg you are right. I've now run the index script.

PR #19, opened 2018-09-07
New check to make sure that that every distinct experiment, epigenome…

thomasmaurel:
… and feature_type combination is linked to a feature_set
Port of the following check: https://github.com/Ensembl/ensj-healthcheck/blob/release/94/src/org/ensembl/healthcheck/testcase/funcgen/ExperimentHasFeatureSet.java
I merged the two SQLs to report combination of column which are not linked to peak_calling.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/ExperimentHasFeatureSet.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/ExperimentHasFeatureSet.pm
@@ -1,0 +1,53 @@
+=head1 LICENSE
+
+Copyright [2018] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/ExperimentHasFeatureSet.pm
```diff
@@ -0,0 +1,53 @@
...
+  GROUPS      => ['funcgen_integrity', 'funcgen_Post_regulatory_build'],
+  DB_TYPES    => ['funcgen'],
+  TABLES      => ['experiment','feature_type','epigenome','peak_calling'],
+};
+
+sub tests {
+  my ($self) = @_;
+  my $desc = "Every distinct experiment, epigenome and feature_type combination is linked to a feature_set";
+  my $sql = q/
+    SELECT DISTINCT ex.experiment_id, ex.epigenome_id, ex.feature_type_id, ep.display_label, ft.name FROM
```
james-monkeyshines: For any rows that are found, because there is not an explicit $diag message, it will print a generic error message, and then the values of these fields. Which wouldn't be helpful, because you would have three arbitrary-looking numbers to start with, and you wouldn't know what columns they related to.

So I think we need a $diag message which has the column names. But this has already come up a couple of times, which makes me think that it'd be good to have a better default diagnostic message, which prints the columns names unless you override it.

thomasmaurel: I see, yes, that make sense now.

PR #21, opened 2018-09-13
Check controlled tables in production database

james-monkeyshines:
This is a replacement for the ProductionMasterTables check, with a modification so that any rows that are in the production db but not in the core, core-like, or variation db are _not_ considered failures.


james-monkeyshines:
Commit history got muddled. Will resubmit on a clean branch,

PR #25, opened 2018-10-09
ENSCORESW-2719: contig can be not null if only one assembly is stored

MatBarba:
Missing condition copied over from the old healthcheck.


james-monkeyshines:
I rolled the code for detecting multiple assemblies into the datacheck sql, since the two statements ended up being very similar. Also, switched this to be the second test, the primary purpose of the datacheck is to check that dna is associated with contigs, so makes sense to have that first.

PR #26, opened 2018-10-09
Properly filter the datacheck to run in the fan runnable

MatBarba:
This is a fix for the following problem:
While running the DbDataChecks pipeline with -parallelize_datachecks 1, the DataCheckFan fails with "X datachecks found: YYY" while listing the sole datacheck name provided (YYY here).

There may be a better way to fix this.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Pipeline/DataCheckFan.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Pipeline/DataCheckFan.pm
@@ -35,17 +35,22 @@ sub run {
   my $manager          = $self->param_required('manager');
   my $datacheck_params = $self->param_required('datacheck_params');
 
+  # Filter datacheck to run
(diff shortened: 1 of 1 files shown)


james-monkeyshines:
This module should never be getting multiple datachecks; the preceding factory module should be doling out one datacheck at a time. But the conf file was setting all the datacheck filtering parameters as well, which I think was messing things up. I've checked in a fix for that to the master branch (https://github.com/Ensembl/ensembl-datacheck/commit/d58aa12), and have been unable to replicate the problem. @MatBarba, please check if that resolves the issue with your use case.

MatBarba:
@james-monkeyshines I can confirm that this change fixed my case. Thanks!

james-monkeyshines:
Jolly good, thanks for reporting the problem (and for suggesting a fix, even though I didn't use it).

PR #30, opened 2018-10-15
Test should pass for 0 or 1 rows

MatBarba:
It looks like the present test doesn't accept any contig with a name? I think it should pass if it gets one row too.


james-monkeyshines:
You are correct, the SQL was flawed. I went for something a bit less complicated, for ease of readability/maintenance, and to take advantage of diagnostics provided by the is_rows_zero method.
Tested on metazoa and vertebrates, and negative controls.

MatBarba:
Looks good to me, thanks!

PR #34, opened 2018-10-23
Feature/new datachecks

james-monkeyshines:


james-monkeyshines:
Test coverage has gone down because the new function 'denormalized' is not tested. This requires a test variation db, which we haven't got yet.

PR #37, opened 2018-11-01
New datachecks, mostly related to seq_regions

james-monkeyshines:
Also includes fixes to problems I noticed with collection dbs using up all the connections.


thomasmaurel:
I think coverage of 93.431% is still good ;-) so merging this.

PR #38, opened 2018-11-06
Feature/advisory dc

amonida:


lib/Bio/EnsEMBL/DataCheck/Checks/DisplaybleGenes.pm
```diff
@@ -0,0 +1,69 @@
...
+
+    my $sql = 'SELECT COUNT(*) FROM analysis_description';
+
+    if (! sql_count($self->dba, $sql) ) {
+      return (1, 'No analysis_description');
+    }
+  }
+
+  my $sql = q/
+    SELECT COUNT(gene_id)
```
james-monkeyshines: Using count is fine, but I just wanted to point out that you get bonus diagnostics if you do a non-count select. For example, if the test fails, you probably want to know the analysis it failed for. If you did `SELECT DISTINCT logic_name`, the offending logic_names will be reported in the test failure.
The disadvantage is that this query will take a bit longer than the count, but usually it's a matter of milliseconds, so that's not normally an issue.
If you do it this was, it's best to add a diagnostic message; by default it'll say 

> Unexpected data: (logic_name) = (ensembl_genomes)

but you can replace that first bit with something more helpful, like "Undisplayed analysis", by setting that as an additional `$diag` parameter, and passing it as the last option to `is_rows_zero`.
(See https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/SequenceLevel.pm for an example)

lib/Bio/EnsEMBL/DataCheck/Checks/DisplaybleGenes.pm
```diff
@@ -0,0 +1,69 @@
...
+      FROM analysis, gene, analysis_description
+     WHERE gene.analysis_id = analysis.analysis_id
+       AND analysis.analysis_id = analysis_description.analysis_id
+       AND analysis_description.web_data is not NULL
```
james-monkeyshines: The logic here is incorrect; we want to find instances of web_data being NULL, since that is problematic. We also want to find either web_data NULL or displayable = 0, so would need an OR between them; but that's not applicable if you do as I suggest above and split into two tests.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplaybleGenes.pm
```diff
@@ -0,0 +1,69 @@
...
+       AND analysis_description.displayable = 0
+  /;
+
+  my $desc = 'All genes are displayed correctly.';
```
james-monkeyshines: My recommended format is to have the `$desc` before the `$sql`. That way the description acts as a comment for the query.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplaybleGenes.pm
```diff
@@ -0,0 +1,69 @@
...
+  GROUPS         => ['core_handover'],
+  DATACHECK_TYPE => 'advisory',
+  DB_TYPES       => ['core', 'otherfeatures', 'rnaseq', 'cdna'],
+  TABLES         => ['gene', 'analysis', 'analysis_description']
+};
+
+sub tests {
+  my ($self) = @_;
+
+  sub skip_tests {
```
james-monkeyshines: I don't think this is required; we should pretty much always have analysis_description rows. So we're doing a redundant count, since we will almost never need to skip.
In any case, this sub should not be within the `tests` sub, it should just be alongside it.

james-monkeyshines: `skip_tests` is no longer within `tests`, but my main point here was that it shouldn't be there at all...

PR #40, opened 2018-11-09
New datachecks for karyotype, metadata and analyses

james-monkeyshines:
As an aside, note that using the API enabled us to side-step all the horrible LRG and patch exceptions in the HC queries...


lib/Bio/EnsEMBL/DataCheck/Checks/ChromosomesAnnotated.pm
```diff
@@ -0,0 +1,73 @@
...
+  NAME           => 'ChromosomesAnnotated',
+  DESCRIPTION    => 'Check whether chromosomal seq_regions have appropriate attribute.',
+  GROUPS         => ['assembly'],
+  DB_TYPES       => ['core'],
+  TABLES         => ['attrib_type', 'coord_system', 'seq_region', 'seq_region_attrib']
+};
+
+sub skip_tests {
+  my ($self) = @_;
+  my $species_id = $self->dba->species_id;
```
amonida: Maybe consider using the DC species method just like in the AttribValues DC?

james-monkeyshines: AttribValues is restricting the application of the datacheck to a subset of species. In this case I use species_id to make sure the query (and subsequent skipping) is restricted to one species at a time for collection dbs.

PR #43, opened 2018-11-14
New test/alt allele

marcoooo:
New data checks for AltAllele and AltAllele group members check.


james-monkeyshines:
@amonida As our in-house expert on alt_alleles, your evaluation of the code would be much appreciated...

amonida:
All good!

PR #47, opened 2018-11-21
FeaturePosition HC2DC

marcoooo:
FeaturePosition HC migration.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
@@ -30,26 +30,19 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
     NAME        => 'FeaturePosition',
     DESCRIPTION => 'Checks if features lie within bounds of seq_region i.e. start >=0 and end <= seq_region length.',
-    GROUPS      => [ 'corelike_handover' ],
-    DB_TYPES    => [ 'funcgen' ]
+    GROUPS      => [ 'funcgen_handover' ],
+    DB_TYPES    => [ 'funcgen' ],
+    TABLES      => [ "peak", "regulatory_feature", "motif_feature", "external_feature", "mirna_target_feature" ]
 };
 
 sub tests {
   my ($self) = @_;
 
   my $dna_dba = $self->get_dna_dba();
-  my $mapper = sub {
-    my ($row, $value) = @_;
-    my %row = (
-        region_name   => $$row[1],
-        region_length => $$row[2]
-    );
-    return \%row;
-  };
+
   # process with db is not master
   my $sql_regions = q/
     SELECT  seq_region_id,
-            seq_region.name,
             `length`
     FROM    seq_region
     JOIN    seq_region_attrib USING (seq_region_id)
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
```diff
@@ -0,0 +1,108 @@
...
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Utils qw/sql_count/;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+    NAME        => 'FeaturePosition',
+    DESCRIPTION => 'Checks if features lie within bounds of seq_region i.e. start >=0 and end <= seq_region length.',
+    GROUPS      => [ 'corelike_handover' ],
```
james-monkeyshines: This isn't a corelike datacheck, 'funcgen_handover' would be more appropriate.

lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
```diff
@@ -0,0 +1,108 @@
...
+    DB_TYPES    => [ 'funcgen' ]
+};
```
james-monkeyshines: Could also have TABLES defined here; this is a potentially long-running datacheck, so it's useful to specify the list of `@feature_tables` here, so that we only re-run if anything in those changes.

lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
```diff
@@ -0,0 +1,108 @@
...
+
+sub tests {
+  my ($self) = @_;
+
+  my $dna_dba = $self->get_dna_dba();
+  my $mapper = sub {
```
james-monkeyshines: You could do a core API call here, instead of the SQL, to get the list of seq_regions. E.g. 
`my $sa = $dna_dba->get_adaptor('Slice');`
`my $slices = $sa->fetch_all('toplevel');`
Then can iterate over the `$slices` array_ref, getting the required data from the `get_seq_region_id` and `seq_region_length` methods.

But... that could be quite slow. So SQL may be the best, but it could be simplified some. The `$region_name` is not used anywhere, so you could just retrieve seq_region_id and length, and do an `execute_into_hash` without the `$mapper` sub.

lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
```diff
@@ -0,0 +1,108 @@
...
+  my @feature_tables = ("peak", "regulatory_feature", "motif_feature", "external_feature", "mirna_target_feature");
+
+  foreach my $feature_table (@feature_tables) {
+    my $desc_feature_ok = "No " . $feature_table . "s exceeding seq_region bounds";
+    my $desc_feature_pos = "";
+    my $feature_count = 0;
+    foreach my $region_id (keys %regions) {
+      my %region_info = %{$regions{$region_id}};
+      # print 'region_id' . $region_id . "\n";
+      my $region_name = $region_info{'region_name'};
```
james-monkeyshines: `$region_name` variable is not used anywhere.

lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
```diff
@@ -0,0 +1,108 @@
...
+      my $region_length = $region_info{'region_length'};
+      # print $region_info{'region_name'} . "\n";
+      # print $region_info{'region_length'} . "\n";
+      # print $feature_table . "\n";
+      my $feature_pos_sql;
+      if ($region_name eq "regulatory_feature") {
```
james-monkeyshines: Wrong variable name; should be `$feature_table` rather than `$region_name`.

lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
```diff
@@ -0,0 +1,108 @@
...
+          FROM    $feature_table
+          WHERE   seq_region_id= $region_id
+          AND     (seq_region_start <= 0 OR seq_region_end > $region_length)
+          /;
+      }
+      $feature_count += sql_count($self->dba, $feature_pos_sql);
+    }
+    $desc_feature_pos = "Found " . $feature_count . " for " . $feature_table . "s exceeding seq_region bounds";
+    if ($feature_count > 0) {
+      fail($desc_feature_pos);
```
james-monkeyshines: In general, `fail` methods should be avoided; it's most often a sign that the test hasn't been formulated correctly. Instead of this conditional, you can do
`is($feature_count, 0, $desc_feature_ok);`
That also removes the need for the explicit failure message (`$desc_feature_pos`), because the test will automatically report something like 'Expected 0, Found 312'.

lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
```diff
@@ -0,0 +1,108 @@
...
+    }
+    else {
+      ok($desc_feature_ok);
```
james-monkeyshines: This isn't doing what you want; it's testing whether the `$desc_feature_ok` string returns a true value. Using `pass` instead of `ok` would do what you expect. But using the `pass` method should be avoided, see previous comment on `fail`.

lib/Bio/EnsEMBL/DataCheck/Checks/SampleRegulatoryFeatureExists.pm
```diff
@@ -24,7 +24,7 @@ use strict;
 use Moose;
 use Test::More;
 use Bio::EnsEMBL::DataCheck::Test::DataCheck;
-use Bio::EnsEMBL::DataCheck::Utils qw/sql_count/;
+use Bio::EnsEMBL::DataCheck::Utils ``;
```
james-monkeyshines: This seems like a mistake?

lib/Bio/EnsEMBL/DataCheck/Checks/FeaturePosition.pm
marcoooo: Unfortunately `$region_name` is used to test if this is `regulatory_feature`, then the query is sligthly different.

PR #50, opened 2018-11-22
Adding a fairly dumb HC developed during the retreat

dbolser-ebi:
Please feel free to discard this James. May be useful as an example of something super trivial. Come to think of it, this may have just been copypasta from the examples on the slides.


james-monkeyshines:
Thanks Dan. I think this is a worthwhile check, but it doesn't go far enough, there are other repeat sources to check for, and it needs to be made collection-db-aware. It's easier for me to create a new datacheck from scratch, for which I've created myself a ticket (https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-2701)

PR #48, opened 2018-11-22
New test/exon strand order

marcoooo:
ExonStrandOrder

PR #51, opened 2018-11-27
New test/exons boundaries

marcoooo:
This is the resulting merged from ExonStrandOrder (https://github.com/Ensembl/ensembl-datacheck/pull/48) and ExonTranscriptStartEnd HealthChecks. 
**Only one of the PR should be merged**


.gitignore
```diff
@@ -1,2 +1,4 @@
 MultiTestDB.conf
-.idea
\ No newline at end of file
+.idea
+/_Deparsed_XSubs.pm
+E
```
james-monkeyshines: I don't object to additions to .gitignore, but this looks a bit like a mistake?

lib/Bio/EnsEMBL/DataCheck/Checks/ExonsBoundaries.pm
```diff
@@ -0,0 +1,163 @@
...
+        $lastExonEnd = $$strand{exon_end};
+        $lastExonID = $$strand{exon_id};
+        $lastExonRank = $$strand{exon_rank};
+        $minExonStart = $$strand{transcript_start};
+        $maxExonEnd = $$strand{transcript_end};
+        $lastTranscriptEnd = $$strand{transcript_end};
+        $lastTranscriptStart = $$strand{transcript_start};
+      }
+      else {
+        if ($$strand{exon_strand} != $lastExonStrand) {
```
james-monkeyshines: This test isn't necessary. We have a test to check whether exon and transcript are different. By definition, for a multi-exon transcript, if two exons have different strands, then one of them will be different to the transcript; so checking whether exons have different strands is redundant.

lib/Bio/EnsEMBL/DataCheck/Checks/ExonsBoundaries.pm
```diff
@@ -0,0 +1,163 @@
...
+    NAME        => 'ExonsBoundaries',
+    DESCRIPTION => 'Checks exon boundaries in their transcript (same strand / correct order / min start / max end)',
+    DB_TYPES    => [ 'core' ],
+    GROUPS      => [ 'core_handover' ],
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $mapper = sub {
```
james-monkeyshines: I'm not a fan of this approach. It necessitates the creation of lots of variables (most of which are counter-intuitively initialised to negative values) which then need to be tracked and incremented. I appreciate that this is how it was done in the HC, and that it works; it's just difficult to understand and maintain. Another disadvantage is that the logic for several tests is all bundled up together; if we wanted to split the functionality up, it'd be tricky to disentangle.

james-monkeyshines: My suggestion would be to copy the functionality from the https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/GeneTranscriptBounds.pm datacheck. This does for genes and transcripts most of what needs to be done for transcripts and exons: there's a test for checking seq_region and strand consistency, and that the extremities match up.

That just leaves the exon overlap functionality. I imagine that could be done with a pretty hairy SQL query; but the existing approach here is probably better. Without all the other functionality, the code to check for overlaps is greatly simplified.

.gitignore
marcoooo: for the E yes.

PR #52, opened 2018-11-28
Add a datacheck for biotype table in core databases

mkszuba:
See ENSCORESW-2824. Part of the core_handover group.

It:
 * checks if the biotype table is in sync with the master copy in the production database;
 * checks if all biotypes used in the gene and transcript tables are listed in the biotype table;
 * one needs to use the parameter 'old_server_uri' to point the datacheck to the production database, which is not intuitive. Therefore, if that option isn't specified tell the users this is what they should use.


lib/Bio/EnsEMBL/DataCheck/Checks/CoreBiotypes.pm
```diff
@@ -0,0 +1,103 @@
...
+# database to be checked and a Production database containing the
+# master copy of the biotype table. The former is specified the
+# standard way, the latter take advantage of - and possibly slightly
+# abuses - the option -old_server_uri.
+sub tests {
+  my ($self) = @_;
+
+  my $master_dba;
+  try {
+    $master_dba = $self->get_old_dba();
```
james-monkeyshines: This is not what `get_old_dba` is for; you need to supply a registry file, and then use `get_dba`, _a la_ https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCore.pm#L41

lib/Bio/EnsEMBL/DataCheck/Checks/CoreBiotypes.pm
```diff
@@ -0,0 +1,103 @@
...
+    -SQL => $sql_whole_biotype_table
+  );
+
+  my $master_biotype_table = $master_sql_helper->execute(
+    -SQL => $sql_whole_biotype_table
+  );
+
+  # Reminder: this *will* fail if results from both tables are not
+  # ordered in the same way
+  cmp_deeply( $this_biotype_table, $master_biotype_table,
```
james-monkeyshines: I've steered clear of Test::Deep methods in datachecks; as your comment notes, they sometimes have particular requirements, but the killler thing for me is the poor diagnostics. The failure messages give you no indication about what rows have caused the problem. I've tackled this sort of thing by loading up some hashes and doing an `is_deeply` (https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCore.pm#L72)

lib/Bio/EnsEMBL/DataCheck/Checks/CoreBiotypes.pm
```diff
@@ -0,0 +1,103 @@
...
+              'biotype table in sync with master' );
+
+  foreach my $object_type ( 'gene', 'transcript' ) {
+    subtest "definitions of $object_type biotypes" => sub {
+
+      my $biotypes_from_objects = $sql_helper->execute_simple(
+        -SQL => "SELECT DISTINCT biotype FROM $object_type"
```
james-monkeyshines: This query doesn't have a species_id constraint; if this datacheck was run on a collection database with, say, 200 species, it would be run 200 times. Probably not terrible in terms of performance due to query caching, but not how things are meant to work (and the output would be 200 times the size it should be). You _could_ set the PER_DB property to 1, but it's generally better to add a constraint; if there is a problem, it allows you to see exactly which species are problematic, and it makes it easier to parallelise when run as part of a hive pipeline.

james-monkeyshines: E.g. https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/GeneBiotypes.pm#L60

lib/Bio/EnsEMBL/DataCheck/Checks/CoreBiotypes.pm
```diff
@@ -0,0 +1,103 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CoreBiotypes;
```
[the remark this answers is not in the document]
james-monkeyshines: No worries; we are tracking the HCs that we convert to datachecks with an epic in Jira  (https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-1995). The ProductionBiotypes HC became the GeneBiotypes datacheck, which I augmented based on a request from Dan B (https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-2082).

PR #69, opened 2019-01-18
New datachecks

james-monkeyshines:
Analysis format, descriptions (via ForeignKeys update), polyploid attribs, and align feature links to external dbs.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/AnalysisFormat.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/AnalysisFormat.pm
@@ -49,7 +49,9 @@ sub tests {
   my $desc_2 = 'Created date is not all zeroes';
   my $sql_2  = q/
     SELECT logic_name FROM analysis
-    WHERE created = '0000-00-00 00:00:00'
+    WHERE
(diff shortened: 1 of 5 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/AnalysisFormat.pm
```diff
@@ -0,0 +1,57 @@
...
+  my $sql_1  = q/
+    SELECT logic_name FROM analysis
+    WHERE BINARY logic_name <> lower(logic_name)
+  /;
+  is_rows_zero($self->dba, $sql_1, $desc_1);
+
+  my $desc_2 = 'Created date is not all zeroes';
+  my $sql_2  = q/
+    SELECT logic_name FROM analysis
+    WHERE created = '0000-00-00 00:00:00'
```
marcoooo: You should test for 1970-00-00 00:00:00 as well, as newer Mysql versions doesn't even allow 0000... value anyway.

james-monkeyshines: Thanks for the tip, done.

PR #70, opened 2019-01-23
Feature/amonida/dc

amonida:


lib/Bio/EnsEMBL/DataCheck/Checks/MTCodonTable.pm
```diff
@@ -0,0 +1,66 @@
...
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::MTCodonTable;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Utils qw/sql_count/;
```
james-monkeyshines: This isn't used anywhere...

lib/Bio/EnsEMBL/DataCheck/Checks/MTCodonTable.pm
```diff
@@ -0,0 +1,66 @@
...
+
+
+use constant {
+  NAME           => 'MTCodonTable',
+  DESCRIPTION    => 'MT seq region had associated seq_region attribute '
+    . 'and correct codon table',
+  GROUPS         => ['core_handover'],
+  DATACHECK_TYPE => 'critical',
+  DB_TYPES       => ['cdna', 'core', 'funcgen', 'otherfeatures', 'rnaseq', 'variation'],
+  TABLES         => ['seq_region', 'seq_region_attrib']
```
james-monkeyshines: There are tables that are in the queries which aren't listed here; need to be comprehensive, attrib_type is unlikely to change, but if it does we could miss out on re-running the datacheck when we need to.

lib/Bio/EnsEMBL/DataCheck/Checks/MTCodonTable.pm
```diff
@@ -0,0 +1,66 @@
...
+
+sub tests {
+  my ($self) = @_;
+
+  my $desc_1 = 'MT seq_regions have associated seq_region_attrib '
+    . 'using correct codon table';
+  my $sql_1 = q/
+      SELECT COUNT(seq_region_id)
+        FROM seq_region
+       WHERE name like '%MT%'
```
james-monkeyshines: This query is not "collection-safe" (cf https://www.ebi.ac.uk/seqdb/confluence/display/GTI/Everything+you+wanted+to+know+about+datachecks%2C+but+were+too+afraid+to+ask#Everythingyouwantedtoknowaboutdatachecks,butweretooafraidtoask-Collectiondatabases).
It needs to link to coord_system and filter on species_id.
Also '%MT%' is perhaps too permissive; it's not too much of a stretch to imagine a non-MT seq that just has 'MT' in the name some where. Maybe '%MT'?

james-monkeyshines: I'm not sure comparing counts is the right thing to do here. I know that's what the HC did, but why not replace the two queries with a single one that returns a list of seq_region_ids and names where the name matches MT, but the attrib is missing, _a la_ https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionTopLevel.pm#L59

james-monkeyshines: Or, one step further, you could use the API, something like https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/ChromosomesAnnotated.pm#L56

lib/Bio/EnsEMBL/DataCheck/Checks/MTCodonTable.pm
```diff
@@ -0,0 +1,66 @@
...
+    /;
+
+  my $sql_2 = q/
+    SELECT COUNT(seq_region_attrib.seq_region_id)
+      FROM attrib_type
+INNER JOIN seq_region_attrib USING (attrib_type_id)
+INNER JOIN seq_region USING (seq_region_id)
+     WHERE attrib_type.code = 'codon_table'
+       AND seq_region_attrib.value = '2'
```
james-monkeyshines: This won't work for non-vertebrates. They have different translation tables: https://www.ncbi.nlm.nih.gov/Taxonomy/Utils/wprintgc.cgi.
Checking for the correct one is overkill, I think it's enough that a value exists. If the code is wrong, it's highly likely that MT proteins will have internal stop codons, which'll get flagged up elsewhere.

PR #74, opened 2019-01-29
Improving "Previous Version" functionality

james-monkeyshines:
Rejig the way that DBAdaptors for previous releases are retrieved. If an explicit db name is not given, look one up in the metadata db. Query old database to get species_id, as that may not be the same between releases.
Add tests for retrieving DBAdaptor for previous release. Requires a test metadata database.


james-monkeyshines:
> Should we not rename the DC to CompareBiotypeGrou[Counts] then ?

Yes. Change submitted.

PR #77, opened 2019-02-07
"Compare previous" datachecks

james-monkeyshines:
Two new variation datachecks, tidying up a few other "Compare" datachecks, and making the behaviour of one of the testing methods more intuitive/sensible.


thomasmaurel:
Looks good to me.

PR #86, opened 2019-03-07
New datachecks for variation 

dglemos:


lib/Bio/EnsEMBL/DataCheck/Checks/Population.pm
```diff
@@ -0,0 +1,75 @@
...
+  my $diag_length = 'Population has no stored size'; 
+  my $sql_length = qq/
+      SELECT p.population_id
+      FROM population p, sample_population sp
+      WHERE p.population_id = sp.population_id
+      AND p.size is NULL
+  /;
+  is_rows_zero($self->dba, $sql_length, $desc_length, $diag_length); 
+
+  my $species = $self->dba->species;
```
james-monkeyshines: It's better to use `$self->species` instead. Because datachecks can be invoked in a variety of ways it is possible (albeit uncommon) to run them such that `dba->species` gets set to the string "DEFAULT". This is avoided if you use the datacheck object's `species` method.

lib/Bio/EnsEMBL/DataCheck/Checks/Phenotype.pm
```diff
@@ -0,0 +1,48 @@
...
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'Phenotype',
```
james-monkeyshines: This is a rather ambiguous name; I'd suggest something like "PhenotypeDescriptionMissing".

lib/Bio/EnsEMBL/DataCheck/Checks/Population.pm
```diff
@@ -0,0 +1,75 @@
...
+
+sub tests {
+  my ($self) = @_;
+
+  my $desc_length = 'Population size'; 
+  my $diag_length = 'Population has no stored size'; 
+  my $sql_length = qq/
+      SELECT p.population_id
+      FROM population p, sample_population sp
+      WHERE p.population_id = sp.population_id
```
james-monkeyshines: It's somewhat a matter of style, so feel free to ignore this one, but... I always find it easier to read a SQL query if the join conditions are separated out from filter conditions. For example, I would write this as:
`SELECT p.population_id`
`FROM population p `
`INNER JOIN sample_population sp USING (population_id)`
`WHERE p.size IS NULL`

lib/Bio/EnsEMBL/DataCheck/Checks/Source.pm
```diff
@@ -0,0 +1,63 @@
...
+  my $sql_url = qq/
+      SELECT source_id
+      FROM source
+      WHERE url NOT REGEXP '^http[s]?:\/\/.*';    
+  /;
+  is_rows_zero($self->dba, $sql_url, $desc_url, $diag_url);
+
+  my $desc_dup = 'Source name is not duplicated';
+  my $diag_dup = 'Source name is duplicated';
+  my $sql_dup = qq/
```
james-monkeyshines: This is fine, but it seems like you could have a unique index on the source.name column instead?

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -421,4 +422,36 @@ sub denormalised {
...
+by calling B<is_rows_zero>. Which means, if the number of 
+rows is zero the test will pass. The C<$id> will be used for 
+a C<SELECT> statement to count the rows that are returned and print 
+them as a diagnostic message. 
+
+=back
+
+=cut
+
+sub no_missing_value {
```
james-monkeyshines: The function looks good, but I'm not a big fan of the name; it's sort of double-negative that takes a few beats to work out what it is testing. Perhaps `has_values` instead? I don't love that either, so feel free to argue/suggest something else.

lib/Bio/EnsEMBL/DataCheck/Checks/SourceAdvisory.pm
```diff
@@ -0,0 +1,71 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::SourceAdvisory;
```
james-monkeyshines: Having 'Advisory' in the name doesn't feel quite right, but I don't have a better suggestion. So this is fine, but if anything occurs to me I'll propose a change at a later date...

dglemos: I agree 'Advisory' is not a good idea but as it's checking different columns and values, I couldn't find a better description.

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
dglemos: I tried to follow the same logic as 'is_rows_zero' - if number of rows zero the test pass. So I thought the name should say 'no_missing_value' - if there are no missing values the test pass. If 'no_missing_value' is too confusing, could it be 'missing_value'? The name 'has_values' is also another possibility as you suggested.

lib/Bio/EnsEMBL/DataCheck/Checks/Source.pm
dglemos: We'll look into that but I'll leave this check as it is for now.

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
james-monkeyshines: Thanks, I appreciate your arguments, in the end I decided to go with 'has_data'.

PR #87, opened 2019-03-08
Add check for TranscriptVariation

helensch:
Converstion of HealthCheck TranscriptVariation.java
ENSVAR-1621

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/TranscriptVariation.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/TranscriptVariation.pm
@@ -1,0 +1,72 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
+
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/TranscriptVariation.pm
```diff
@@ -0,0 +1,72 @@
...
+  
+  my $desc_2 = 'Consequence type is not missing';
+  my $sql_2 = qq/
+      SELECT COUNT(*) 
+      FROM transcript_variation 
+      WHERE consequence_types = ''
+      OR consequence_types = 'NULL'
+      OR consequence_types IS NULL
+  /;
+  is_rows_zero($self->dba, $sql_2, $desc_2);
```
james-monkeyshines: PR #86 includes a new method which could be used here once that gets merged (currently called 'no_missing_values', but I suggested changing it to something else, maybe 'has_values').

PR #88, opened 2019-03-12
Variation checks for Individual type and variation class

dglemos:


lib/Bio/EnsEMBL/DataCheck/Checks/IndividualType.pm
```diff
@@ -0,0 +1,88 @@
...
+  DESCRIPTION => 'Individuals have the correct type for each species',
+  GROUPS      => ['variation_import'],
+  DB_TYPES    => ['variation'],
+  TABLES      => ['individual'],
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $species = $self->dba->species;
```
james-monkeyshines: As mentioned in the other PR, this should be `$self->species`, rather than `$self->dba-species`.
I appreciate that this is not at all obvious, I'll add a bit about it on confluence.

PR #94, opened 2019-03-22
added critical datacheck DisplayableSampleGene

brunocontrerasmoreira:
added critical datacheck DisplayableSampleGene to avoid issues such as https://github.com/Ensembl/staging-patches/pull/251

Created with:

$ perl -I lib scripts/create_datacheck.pl -n DisplayableSampleGene -d 'Sample gene is displayable and has web_data attached to its analysis' -g core_handover -db core -t meta,gene,analysis,analysis_description

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleGene.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleGene.pm
@@ -1,0 +1,75 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::DisplayableSampleGene;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'DisplayableSampleGene',
+  DESCRIPTION => 'Sample genes is displayable and has web_data attached to its analysis',
+  GROUPS      => ['core_handover'],
+  DB_TYPES    => ['core'],
+  TABLES      => ['analysis', 'analysis_description', 'gene', 'meta']
+};
+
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleGene.pm
```diff
@@ -0,0 +1,75 @@
...
+
+use Moose;
+use Test::More;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'DisplayableSampleGene',
+  DESCRIPTION => 'Sample gene is displayable and has web_data attached to its analysis',
+  GROUPS      => ['core_handover'],
```
james-monkeyshines: There is no 'handover' group, the 'core' group is equivalent. Also makes sense to add this to the 'geneset' group.
```suggestion
  GROUPS      => ['core', 'geneset'],
```

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleGene.pm
```diff
@@ -0,0 +1,75 @@
...
+  my $diag_1 = 'Undisplayed analysis';
+  my $sql_1  = qq/
+      SELECT gene_id 
+        FROM gene g
+  INNER JOIN meta m 
+          ON g.stable_id = m.meta_value 
+         AND m.meta_key = 'sample.gene_param'
+  INNER JOIN analysis a ON g.analysis_id = a.analysis_id
+  INNER JOIN analysis_description ad 
+          ON g.analysis_id = ad.analysis_id AND ad.displayable = 1
```
james-monkeyshines: The SQL will return the gene id if the track is displayable; but the test expects zero rows to be returned, in order to pass. So either the condition should be `ad.displayable = 0` or the test should be `is_rows_nonzero`.
But, in order to degrade gracefully when a sample gene does not exist (vertebrate species may have them defined in text files), the former approach should be used.
```suggestion
          ON g.analysis_id = ad.analysis_id AND ad.displayable = 0
```

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleGene.pm
```diff
@@ -0,0 +1,75 @@
...
+  my $diag_2 = 'web_data is not set';
+  my $sql_2  = qq/
+      SELECT gene_id 
+        FROM gene g
+  INNER JOIN meta m 
+          ON g.stable_id = m.meta_value 
+         AND m.meta_key = 'sample.gene_param'
+  INNER JOIN analysis a ON g.analysis_id = a.analysis_id
+  INNER JOIN analysis_description ad 
+          ON g.analysis_id = ad.analysis_id AND ad.web_data IS NOT NULL
```
james-monkeyshines: Same reasoning as previous comment.
```suggestion
          ON g.analysis_id = ad.analysis_id AND ad.web_data IS NULL
```

brunocontrerasmoreira:
Thanks James, I've accepted your suggested edits

PR #95, opened 2019-03-22
Tests for variation_set and variation_set_structure

dglemos:


lib/Bio/EnsEMBL/DataCheck/Checks/VariationSet.pm
```diff
@@ -0,0 +1,75 @@
...
+  TABLES      => ['variation_set']
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $desc_name = 'Variation set has a name';
+  my $diag_name = 'Variation set name is missing';
+  my $sql_name = q/
+      SELECT count(*)
```
james-monkeyshines: It seems like this could be done more succinctly with the new 'has_data' method (formerly 'no_missing_values')?

dglemos: Yes, I'll change it.

james-monkeyshines: Thanks!

PR #96, opened 2019-03-25
Datacheck for comparing variation feature counts between databases

helensch:
Based on ComparePreviousVariationFeatures healthcheck.
It includes a join to the core seq_region and coord_system tables - should these be included in TABLES?
The assemblies are taken from database names - should the meta database be used instead?


lib/Bio/EnsEMBL/DataCheck/Checks/CompareVariationFeatures.pm
```diff
@@ -0,0 +1,88 @@
...
+    skip 'No old version of database', 1 unless defined $old_dba;
+  
+    # Check the assembly version. Skip if not the same
+    skip 'Different assemblies', 1 
+      unless $self->same_assembly($old_dba->dbc->dbname);
+    
+    my $desc = "Consistent variation feature counts between ".
+               $self->dba->dbc->dbname.' and '.$old_dba->dbc->dbname;
+
+    my ($core_dbname, $old_core_dbname);
```
james-monkeyshines: In order to get the current core db, you can use the `$self->get_dna_db` method. This is a wrapper around the core API's `dnadb` method, which does a bit of extra error-checking/housekeeping.

To get the old core db, you cannot use that method, it only works for `$self->dba`. Instead you can use `$self->get_old_dba(undef, 'core')`. The first parameter is the species (it defaults to that of `$self->dba`), the second is the database group.

With those two core DBAdaptors, you can then retrieve assembly information, and use the API to get top-level seq_regions (relying on `rank=1` for this is generally not a good idea). You then need some code to combine that with a per-seq_region_id query on variation_feature, and then to compare the results between old and new.

So the code will be more complicated, but we can't rely on the core database always being on the same server as the variation one; it's more robust and flexible if we use whatever core is available in the registry. For some pointers on cross database tests, see https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/ForeignKeysMultiDB.pm#L48 and https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/MetaCoord.pm#L122

helensch: Thanks for the pointers. The datacheck has been modified to use the seq region table from the variation database and not the seq region table from the core datase. The ForeignKeysMultipleDB checks that the variation data seq_region IDs and names are consistent with the core database. If this fails, then this check should perhaps not be run. Repeating the same check on core/variation seq region ids in this check seemed like a duplication,

PR #97, opened 2019-03-25
Check variation_set and attrib tables

dglemos:


lib/Bio/EnsEMBL/DataCheck/Checks/VariationSetAttrib.pm
```diff
@@ -0,0 +1,56 @@
...
+  my ($self) = @_;
+
+  my $desc = 'The short name attrib for variation_set_id exists and is defined in attrib table';
+  my $diag = 'The short name attrib for variation_set_id is not defined in attrib table';
+  my $sql = q/
+      SELECT v.variation_set_id
+      FROM variation_set v
+      LEFT JOIN attrib a
+      ON (v.short_name_attrib_id = a.attrib_id)
+      WHERE v.variation_set_id IS NOT NULL
```
james-monkeyshines: The first part of the condition isn't necessary, variation_set_id is an auto_increment/primary key, will always have a value.

dglemos:
Thanks for the comment - v.variation_set_id IS NOT NULL - it's unnecessary. But I'm not sure about this test, won't this be picked up by foreignkeys test?

james-monkeyshines:
This datacheck isn't quite redundant, but it almost is. The short_name_attrib_id column is nullable; the foreign key check would pick up any discrepancies in ids, but it would be happy with null values. This test is effectively disallowing null short_name_attrib_ids (note that attrib.value is not nullable, so the only time that the test would fail, and the foreign keys check wouldn't fail, is if short_name_attrib_id is null). So a schema change could render this redundant, but maybe there's a reason that isn't already in place.

PR #104, opened 2019-03-28
Variation sets and subsets 

dglemos:


lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -40,7 +40,7 @@ our @EXPORT  = qw(
   is_rows cmp_rows is_rows_zero is_rows_nonzero 
   row_totals row_subtotals
   fk denormalized denormalised
-  has_data
+  has_data get_data
```
james-monkeyshines: This is not the right place for such a method. This module is for tests, general purpose methods should go in Utils.pm. But in any case, I don't think this method is required; a range of methods for fetching data already exist in the core API, and those are the recommended approach for datachecks: https://www.ebi.ac.uk/seqdb/confluence/display/GTI/Everything+you+wanted+to+know+about+datachecks%2C+but+were+too+afraid+to+ask#Everythingyouwantedtoknowaboutdatachecks,butweretooafraidtoask-SQLHelpermethods

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
```diff
@@ -0,0 +1,102 @@
...
+    /;
+    my $data = get_data($self->dba, $sql_2);
+    
+    my $test_result;
+    foreach my $element (@$data){
+      my $set1 = @$element[0];
+      my $set2 = @$element[1]; 
+
+      my $data_set1 = $self->get_set($set1);
+      my $related = $set2 ~~ $data_set1;
```
james-monkeyshines: Smart matching makes me a bit twitchy. We require perl 5.14 at a minimum, so it's probably ok, but I think this is sufficiently non-standard to warrant an explanatory comment.

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
```diff
@@ -0,0 +1,102 @@
...
+  
+    skip 'Not running test for variations in superset', 1 unless $result == 1;
+ 
+    my $desc_2 = 'Variations that are in a set are not present in the subset';
+    my $sql_2 = qq/
+        SELECT DISTINCT vs1.variation_set_id, vs2.variation_set_id
+        FROM variation_set_variation vs1
+        JOIN variation_set_variation vs2
+        ON (vs2.variation_id = vs1.variation_id
+        AND vs2.variation_set_id > vs1.variation_set_id) limit 100
```
james-monkeyshines: Why limit to 100? Won't this limit the tests to a subset of data, potentially missing problems?

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
```diff
@@ -0,0 +1,102 @@
...
+      my $data_set1 = $self->get_set($set1);
+      my $related = $set2 ~~ $data_set1;
+      
+      if(!$related){
+        my $data_set2 = $self->get_set($set2);
+        $related = $set1 ~~ $data_set2;
+      }
+      if($related){
+        $test_result = 1; 
+        diag("The variation set ($set1) contains variants that are present in the subset ($set2). Preferably only subset ($set2) should contain those variants");
```
james-monkeyshines: Won't the logic be the other way round sometimes, with $set1 being the subset, depending on whether the first or second smart match result?

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
```diff
@@ -0,0 +1,102 @@
...
+      if(!$related){
+        my $data_set2 = $self->get_set($set2);
+        $related = $set1 ~~ $data_set2;
+      }
+      if($related){
+        $test_result = 1; 
+        diag("The variation set ($set1) contains variants that are present in the subset ($set2). Preferably only subset ($set2) should contain those variants"); 
+      }
+    }
+    ok(!defined($test_result), $desc_2);
```
james-monkeyshines: Rather than just having a flag, I suggest using a variable to count how many failures you get, and then using `is` to compare the value to zero; you get a more informative error message that immediately gives you an idea of the scale of the problem.

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
dglemos: Sorry, that's a mistake.

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
dglemos: Thanks for the suggestion.

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
dglemos: Fixed. Thanks!

lib/Bio/EnsEMBL/DataCheck/Checks/VariationSubset.pm
```diff
@@ -0,0 +1,119 @@
...
+
+package Bio::EnsEMBL::DataCheck::Checks::VariationSubset;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use Data::Dumper;
```
helensch: Is Data::Dumper needed?

dglemos: No, I'll delete it.

PR #112, opened 2019-07-16
Extend input_id field to allow easy submission of jobs via the rest service

james-monkeyshines:


james-monkeyshines:
Travis will fail for now, because it's still configured for perl version 14. I'm trying to fix that on the release/98 branch, so that will get merged in due course.

PR #115, opened 2019-07-26
Remove dead code

wbazant:


james-monkeyshines:
Thanks for spotting this @wbazant, but please could you resubmit it against the master branch, since it does not cause an error, and it makes it a lot easier to merge if we don't have to do it in two release branches as well.

wbazant:
@james-monkeyshines sorry for the incorrectly submitted PR. Do deal with this as you wish: I've left the job, and won't submit any more Ensembl PRs.

PR #116, opened 2019-07-26
Remove the condition that a repeat must start above zero

wbazant:
We have some genomes (ancylostoma_ceylanicum_prjna72583, caenorhabditis_angaria_prjna51225, caenorhabditis_japonica_prjna12591, caenorhabditis_sinica_prjna194557, ditylenchus_dipsaci_prjna498219, schistosoma_bovis_prjna451066 ) that fail this check.

I am not sure why this would be incorrect that a scaffold starts with a repetitive region - in fact I believe that if the repeat is unresolved, and the assembler couldn't go through the junction, the assembly is gonna look like that: it will start with a repeat.

WormBase ParaSite data would like to have this datacheck for making sure analyses got done - and we know some of our assemblies are bad :)

Could this check be moved or removed?

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/RepeatFeatures.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/RepeatFeatures.pm
@@ -81,18 +81,6 @@ sub tests {
       coord_system.species_id = $species_id
   /;
   is_rows_zero($self->dba, $sql_3, $desc_3);
-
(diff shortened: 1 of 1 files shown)


james-monkeyshines:
@wbazant Since everything in Ensembl is 1-based, this is not preventing scaffolds starting with repeat features, the start value in that case will be 1. Any feature that starts at zero is erroneous, likely an off-by-one error, suggesting a potential problem with all of the repeat features from a particular source.

PR #124, opened 2019-08-21
Porting ComparePreviousVersionGOXrefs healthcheck. This compare the G…

thomasmaurel:
…O xrefs counts and group them by souce using the info_type column. This checks for big drops, counts > 1000. The treshold use is 0.1

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/ComparePreviousVersionGOXrefs.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/ComparePreviousVersionGOXrefs.pm
@@ -1,0 +1,72 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::ComparePreviousVersionGOXrefs;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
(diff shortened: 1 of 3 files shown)


thomasmaurel:
@james-monkeyshines: The hc is set to advisory at the moment but it's actually a pass or fail so I guess I should remove this right?

lib/Bio/EnsEMBL/DataCheck/Checks/ComparePreviousVersionGOXrefs.pm
```diff
@@ -0,0 +1,73 @@
...
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'ComparePreviousVersionGOXrefs',
```
james-monkeyshines: I've removed the "PreviousVersion" bit when converting this sort of HC to datachecks, to make it a bit less wordy.
```suggestion
  NAME           => 'CompareProjectedGOXrefs',
```

lib/Bio/EnsEMBL/DataCheck/Checks/ComparePreviousVersionGOXrefs.pm
```diff
@@ -0,0 +1,73 @@
...
+
+    $self->go_xref_counts($old_dba);
+  }
+}
+
+sub go_xref_counts {
+  my ($self, $old_dba) = @_;
+
+  my $minimum_count = 1000;
+  my $threshold = 0.1;
```
james-monkeyshines: I think this needs to be higher - a value of 0.1 means that the new total only has to be 10% of the old total for the test to pass. Presumably the inverse is desired, i.e. 0.9 => 90%

lib/Bio/EnsEMBL/DataCheck/Checks/ComparePreviousVersionGOXrefs.pm
thomasmaurel: Make sense. Done in commit.

lib/Bio/EnsEMBL/DataCheck/Checks/ComparePreviousVersionGOXrefs.pm
thomasmaurel: Yes, that make sense. Sorry, I just copied the treshold from the HC. Fixed in commit.

PR #126, opened 2019-08-21
Converted HC to make sure RefSeq accessions are not predictions

thomasmaurel:

PR #128, opened 2019-08-23
Ported the XrefPrefixes HC which checks that GO, MGI and ZFIN dbprima…

thomasmaurel:
…ry_acc prefix match an expected pattern. I have extended this test to include many xrefs following an expected pattern for both Vert and non-vert. The check has spotted the wrong IDs reported in this ticket: ENSCORESW-3196. I believe that this extended Datacheck now makes the PredictedXrefs check redundant (https://github.com/Ensembl/ensembl-datacheck/pull/126)

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/XrefPrefixes.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/XrefPrefixes.pm
@@ -1,0 +1,110 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::XrefPrefixes;
+
+use warnings;
+use strict;
+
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/XrefPrefixes.pm
```diff
@@ -0,0 +1,111 @@
...
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'XrefPrefixes',
+  DESCRIPTION    => 'Check that xrefs have the correct prefix for their dbprimary_acc',
+  GROUPS         => ['compare_core', 'xref'],
```
james-monkeyshines:
```suggestion
  GROUPS         => ['xref'],
```

lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -1369,6 +1369,16 @@
       "name" : "VariationSubset",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::VariationSubset"
    },
+   "XrefPrefixes" : {
+      "datacheck_type" : "critical",
+      "description" : "Check that xrefs have the correct prefix for their dbprimary_acc",
+      "groups" : [
+         "compare_core",
```
james-monkeyshines:
```suggestion
```

lib/Bio/EnsEMBL/DataCheck/Checks/XrefPrefixes.pm
thomasmaurel: Thanks, fixed in commit.

lib/Bio/EnsEMBL/DataCheck/index.json
thomasmaurel: Thanks, fixed in commit.

PR #129, opened 2019-08-27
Converted the ComparePreviousVersionSynonyms HC. This check compare t…

thomasmaurel:
…he number of external synonyms between two releases for xrefs sources have a count > 500 and using a treshold of 0.78


james-monkeyshines:
Already converted as [CompareSynonym](https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSynonym.pm).

PR #132, opened 2019-08-30
Ported GeneStableIdDisplayXref HC to check that gene display_xref are…

thomasmaurel:
… not set to the stable_id. The DC is now collection friendly.


thomasmaurel:
Merged with TranscriptStableIDDisplayXref HC, see here: https://github.com/Ensembl/ensembl-datacheck/pull/133

PR #134, opened 2019-09-04
Converted HC to check if species inside a collection have protein cod…

thomasmaurel:
…ing genes which are sharing the same display_xref_id. I extended the test to protein coding transcript too as I think it's relevant. It looks like this test only apply to protein coding gene/transcript, when extending to all genes/transcripts I do get back objects with shared display_xref_id but I don't know the reason behind it.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
@@ -32,7 +32,7 @@ use constant {
   DESCRIPTION    => 'Protein coding Gene/Transcript display_xref are not shared between species inside a collection. This can lead to species-specific synonyms being applied to the wrong species',
   GROUPS         => ['xref'],
   DB_TYPES       => ['core'],
-  TABLES         => ['gene', 'transcript', 'seq_region', 'coord_system'],
+  TABLES         => ['gene', 'transcript', 'seq_region', 'coord_system','xref'],
   PER_DB         => 1,
 };
 
@@ -53,13 +53,15 @@ sub shared_display_xref {
   my ($self, $dba, $table) = @_;
 
   my $desc = "No Protein coding $table display_xref shared between species inside a collection database";
-  my $diag = "Protein coding $table display_xref shared between species inside a collection database";
+  my $diag = "Xref accession";
   my $sql  = qq/
     SELECT x.dbprimary_acc FROM
       xref x JOIN 
-      $table tb on (tb.display_xref_id=x.xref_id AND tb.biotype='protein_coding') JOIN
+      $table tb ON tb.display_xref_id=x.xref_id JOIN
       seq_region sr USING (seq_region_id) JOIN
-      coord_system cs USING (coord_system_id) 
+      coord_system cs USING (coord_system_id)
+    WHERE
+      tb.biotype='protein_coding'
     GROUP BY x.dbprimary_acc
     HAVING COUNT(DISTINCT (cs.species_id)) > 1
   /;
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
```diff
@@ -0,0 +1,70 @@
...
+    $self->shared_display_xref($dba,'gene');
+    $self->shared_display_xref($dba,'transcript');
+  }
+}
+
+sub shared_display_xref {
+  my ($self, $dba, $table) = @_;
+
+  my $desc = "No Protein coding $table display_xref shared between species inside a collection database";
+  my $diag = "Protein coding $table display_xref shared between species inside a collection database";
```
james-monkeyshines: Diags aren't really intended to be messages that are just the inverse of the description. The description will be printed in the output immediately before (with an associated negation). The diag is intended more for a short label for the rows that are returned by the SQL statement; in this case that might be "Xref", or "Xref accession".

lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
```diff
@@ -0,0 +1,70 @@
...
+  my $sql  = qq/
+    SELECT x.dbprimary_acc FROM
+      xref x JOIN 
+      $table tb on (tb.display_xref_id=x.xref_id AND tb.biotype='protein_coding') JOIN
```
james-monkeyshines: I think the logic would be clearer if the biotype condition was in a WHERE clause, to distinguish it from the joins. (Incidentally, I think RNA genes don't work with this test because Rfam accessions are used, which are not unique identifiers, it's more like a class of RNA gene, for which there may be many examples.)

lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
```diff
@@ -0,0 +1,70 @@
...
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'SharedDisplayXref',
+  DESCRIPTION    => 'Protein coding Gene/Transcript display_xref are not shared between species inside a collection. This can lead to species-specific synonyms being applied to the wrong species',
+  GROUPS         => ['xref'],
+  DB_TYPES       => ['core'],
+  TABLES         => ['gene', 'transcript', 'seq_region', 'coord_system'],
```
james-monkeyshines:
```suggestion
  TABLES         => ['gene', 'transcript', 'seq_region', 'coord_system', 'xref'],
```

lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
thomasmaurel: Thanks for that. Fixed in commit.

lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
thomasmaurel: Thanks for the explanation. Fixed in commit.

lib/Bio/EnsEMBL/DataCheck/Checks/SharedDisplayXref.pm
thomasmaurel: Fixed in commit.

PR #139, opened 2019-09-10
Ported the ComparePreviousVersionProjectedSynonyms HC to compare proj…

thomasmaurel:
…ected Synonyms between two releases. Added collection support even if I am not sure it's really needed.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CompareProjectedSynonyms.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CompareProjectedSynonyms.pm
@@ -51,7 +51,7 @@ sub tests {
 sub projected_synonyms_counts {
   my ($self, $old_dba) = @_;
 
-  my $threshold = 0.78;
+  my $threshold = 0.80;
 
   my $desc = "Checking Projected Synonyms between ".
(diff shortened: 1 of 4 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CompareProjectedSynonyms.pm
```diff
@@ -0,0 +1,80 @@
...
+    skip 'No old version of database', 1 unless defined $old_dba;
+
+    $self->projected_synonyms_counts($old_dba);
+  }
+}
+
+sub projected_synonyms_counts {
+  my ($self, $old_dba) = @_;
+
+  my $threshold = 0.78;
```
james-monkeyshines: This threshold seems arbitrarily specific. Might be worth looking across all the 'Compare' tests and settle on something consistent - the CompareSynonym and CompareGOXref use 0.70, CompareProjectedGOXrefs has 0.90. Maybe 0.80?

thomasmaurel: Sure, that make perfect sense. I have updated the other Compare* to 0.80.

PR #140, opened 2019-09-10
Ported ComparePreviousVersionProjectedGeneNames HC. The original hc w

thomasmaurel:
…as doing a lot of checks including overall number of projected gene names, lost of sources, change from one source to another, no change in source. We believe with James that it makes more sense to just look at the overall number of projected Gene Names and report big drops. The DC is compatible with collections

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CompareProjectedGeneNames.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CompareProjectedGeneNames.pm
@@ -51,7 +51,7 @@ sub tests {
 sub projected_gene_name_counts {
   my ($self, $old_dba) = @_;
 
-  my $threshold = 0.9;
+  my $threshold = 0.80;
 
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CompareProjectedGeneNames.pm
```diff
@@ -0,0 +1,77 @@
...
+    skip 'No old version of database', 1 unless defined $old_dba;
+
+    $self->projected_gene_name_counts($old_dba);
+  }
+}
+
+sub projected_gene_name_counts {
+  my ($self, $old_dba) = @_;
+
+  my $threshold = 0.9;
```
james-monkeyshines: See comment on #139

thomasmaurel: Updated to 0.80

PR #148, opened 2019-09-19
Update to multiple region update

helensch:
The check that phenotype_feature does not have a single seq region was failing for some metazoa databases. The has been made a separate check and only for vertebrate database


james-monkeyshines:
Thanks, this looks fine, the travis failure is unrelated, but I'll try and get that fixed before merging this.

PR #149, opened 2019-09-19
Set Ensembl release version in travis

james-monkeyshines:
So that appropriate versions of the Ensembl repos are checked out.
The tests are failing because travis is currently checking out `master`, which has a variation schema change that introduces a discrepancy between the database (release 99) and code (release 100).
Something like the code in the ensembl-production `.travis.yml`, to pick up the `TRAVIS_BRANCH` and use that, doesn't work in this case, because candidate branches in PRs do not have the release number, and so still end up using `master`. It probably easier just to include an update to this variable in the SOP for branching at the start of a release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,13 @@ addons:
     - graphviz
 
 before_install:
-    - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl.git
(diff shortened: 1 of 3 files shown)


james-monkeyshines:
I misdiagnosed the problem - there was an update to the variation sql after I had patched the test database, rather than an issue with branch versions. Don't want to hard-code the ensembl release unless we have to, so close this PR.

PR #156, opened 2019-09-27
Release/99

CristiGuijarro:
Inclusion of two new compara-specific datacheck modules:

1. CigarCheck.pm:
- The cigar_line must not have negative numbers in it 
- The cigar_line must not have zeros in it

2. TagCoverageStats.pm
- For each method_link_species_set_id from method_link_species_set_tag:
```
ref_genome_coverage <= ref_genome_length
(ref_matches+ref_mis_matches+ref_insertions) <= (ref_coding_exon_length - ref_uncovered)
ref_covered <= (ref_coding_length - ref_uncovered)
```
- The same for the non_ref_genomes also.
- For each node_id from species_tree_node_tag:
```
genome_coverage <= genome_length
coding_exon_coverage <= coding_exon_length
```
3. Inclusion of these new modules in the index.json file

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
@@ -29,10 +29,10 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME           => 'CigarCheck',
   DESCRIPTION    => 'The cigar_line must not have negative numbers or zeros in it',
-  DATACHECK_TYPE => 'advisory',
+  GROUPS        => ['compara', 'compara_pairwise_alignments'],
+  DATACHECK_TYPE => 'critical',
   DB_TYPES       => ['compara'],
   TABLES         => ['family_member', 'gene_align_member', 'genomic_align', 'homology_member', 'peptide_align_feature'],
-  PER_DB         => 0
 };
 
 sub tests {
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
@@ -29,7 +29,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME           => 'CigarCheck',
   DESCRIPTION    => 'The cigar_line must not have negative numbers or zeros in it',
-  GROUPS        => ['compara', 'compara_pairwise_alignments'],
+  GROUPS         => ['compara', 'compara_pairwise_alignments'],
   DATACHECK_TYPE => 'critical',
   DB_TYPES       => ['compara'],
   TABLES         => ['family_member', 'gene_align_member', 'genomic_align', 'homology_member', 'peptide_align_feature'],
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
@@ -1,0 +1,82 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CigarCheck;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
```diff
@@ -0,0 +1,82 @@
...
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'CigarCheck',
+  DESCRIPTION    => 'The cigar_line must not have negative numbers or zeros in it',
+  DATACHECK_TYPE => 'advisory',
```
muffato: Let's make this test mandatory

lib/Bio/EnsEMBL/DataCheck/Checks/TagCoverageStats.pm
```diff
@@ -0,0 +1,146 @@
...
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'TagCoverageStats',
+  DESCRIPTION    => 'The coverage must not exceed the genome lengths',
+  DATACHECK_TYPE => 'advisory',
```
muffato: Same here

lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -1327,6 +1334,13 @@
...
    },
+   "TagCoverageStats" : {
+      "datacheck_type" : "advisory",
+      "description" : "The coverage must not exceed the genome lengths",
+      "groups" : [],
+      "name" : "TagCoverageStats",
+      "package_name" : "Bio::EnsEMBL::DataCheck::Checks::TagCoverageStats"
+   },
    "TranscriptBounds" : {
       "datacheck_type" : "critical",
```
muffato: Looks like the correct term is "critical", not "mandatory", sorry

CristiGuijarro:
The new  data-checks have been tested on the release 97 compara database which successfully failed these data-checks

lib/Bio/EnsEMBL/DataCheck/Checks/CigarCheck.pm
```diff
@@ -0,0 +1,82 @@
...
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'CigarCheck',
+  DESCRIPTION    => 'The cigar_line must not have negative numbers or zeros in it',
```
james-monkeyshines: There is no `GROUPS` parameter set here. It's not mandatory for a datacheck to belong to a group, but it is expected, and it makes the datachecks much more useful.

Groups can be named anything, but for fundamental tests, which no database should fail (i.e. the set which we will run at handover) the convention is to use the type of database, so 'compara'. In addition, you can specify as many other groups as you like, in order to logically group datachecks. I'm not sure if there's a good group in this case, but the other datacheck in this PR might have something like 'pairwise_alignments'.

You can browse how other teams have grouped things: http://production-services.ensembl.org/app/datachecks-list

As well as keeping things neat, groups can help a lot when embedding datachecks in pipelines. For example, after running the ProteinFeatures pipeline, we run the 'protein_features' group: https://github.com/Ensembl/ensembl-production/blob/release/98/modules/Bio/EnsEMBL/Production/Pipeline/PipeConfig/ProteinFeatures_conf.pm#L909

```suggestion
  DESCRIPTION    => 'The cigar_line must not have negative numbers or zeros in it',
  GROUPS        => ['compara'],
```

muffato: Let's establish the convention, then.
1. `compara` will be listed in every DC. It's the complete list of all DCs that apply to a Compara databae
2. Other groups will be prefixed by `compara_` to make them easier to list and identify, so `compara_pairwise_alignments`, not just `pairwise_alignments`
3. We use the plural form, e.g. `compara_pairwise_alignments` not `compara_pairwise_alignment`

Does it make sense ?

@CristiGuijarro : `compara_pairwise_alignments` should be added to _CigarCheck_ as well because it applies to the pairwise-alignments databases too (amongst other things). The same way that _TagCoverageStats_ applies to other things than pairwise-alignments (multiple alignments, syntenies)

NB: We'll probably add more groups to these two DCs, e.g. `compara_syntenies`, `compara_gene_trees`, etc, when we add other DCs and start defining extra groups

james-monkeyshines: Sounds sensible to me.

james-monkeyshines:
Keeping the index.json file in sync is a bit of a chore. After getting bitten with this for the umpteenth time, I created a git hook which runs after every commit - it executes the test for the index, and if that fails it updates it and automatically adds it to the commit. Further details: https://github.com/Ensembl/ensembl-datacheck/blob/master/README.md

I find it useful, but one thing you have to watch out for is that the script to update the index will pick up un-added files too, so you can end up with an index entry for something that hasn't been added to the repo yet. If I have several commits to do in a row, then, I pass the `--no-verify` flag to all but the last one.

muffato:
Thank you !

PR #158, opened 2019-09-30
Won't always have a DBA, so make the parsing of its parameters conditional

james-monkeyshines:
Fixes bug in previous bug fix...


james-monkeyshines:
I'll add a test case to check for the parameters before I merge this...

PR #161, opened 2019-10-21
Added script to create JIRA tickets from failed datachecks

JAlvarezJarreta:
See [ENSCOMPARASW-2696](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-2696).

### DIFF
--- a/scripts/create_datacheck_tickets.pl
+++ b/scripts/create_datacheck_tickets.pl
@@ -1,0 +1,172 @@
+#!/usr/bin/env perl
+
+=head1 LICENSE
+Copyright [1999-2015] Wellcome Trust Sanger Institute and the EMBL-European Bioinformatics Institute
+Copyright [2016-2019] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+=cut
+
+=head2 DESCRIPTION
+
+This script parses the output from Ensembl's Datachecks and creates JIRA tickets
+for each failure.
+
+=head1 SYNOPSIS
+
+perl create_datacheck_tickets.pl [options]
+
+=head1 OPTIONS
+
+=over 8
+
+=item B<-i[nput]> <datacheck_TAP_file>
+
+Mandatory. Path to a datacheck output TAP file.
+
+=item B<-r[elease]> <release>
+
+Mandatory. Ensembl release version.
+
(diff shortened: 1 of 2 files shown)


scripts/create_datacheck_tickets.pl
```diff
@@ -0,0 +1,172 @@
...
+        description => $testcase_failures->{$testcase},
+    };
+    push(@json_subtasks, $failure_subtask_json);
+}
+# Add subtasks to the initial ticket
+$dc_task_json_ticket->[0]->{subtasks} = \@json_subtasks;
+# Create all JIRA tickets
+my $dc_task_keys = $jira_adaptor->create_tickets(
+    -JSON_OBJ         => $dc_task_json_ticket,
+    -DEFAULT_PRIORITY => 'Blocker',
```
james-monkeyshines: It'd be useful if this was a script parameter, with this as the default value (since some datachecks might be advisory, and therefore not necessarily blocking).

scripts/create_datacheck_tickets.pl
```diff
@@ -0,0 +1,172 @@
...
+die pod2usage(1) if (! ($dc_file && $release && $project));
+# Get file absolute path and basename
+my $dc_abs_path = abs_path($dc_file);
+my $dc_basename = fileparse($dc_abs_path, qr{\.[a-zA-Z0-9_]+$});
+# Get timestamp that will be included in the summary of each JIRA ticket
+my $timestamp = strftime("%d-%m-%Y %H:%M:%S", localtime time);
+# Get a new Utils::JIRA object to create the tickets for the given division and
+# release
+my $jira_adaptor = Bio::EnsEMBL::Compara::Utils::JIRA->new(
+    -USER     => $user,
```
james-monkeyshines: This instantiation fails, because there is a hard-coded list of acceptable usernames in Bio::EnsEMBL::Compara::Utils::JIRA, line 329

scripts/create_datacheck_tickets.pl
```diff
@@ -0,0 +1,172 @@
...
+=cut
+
+=head2 DESCRIPTION
+
+This script parses the output from Ensembl's Datachecks and creates JIRA tickets
+for each failure.
+
+=head1 SYNOPSIS
+
+perl create_datacheck_tickets.pl [options]
```
james-monkeyshines:
```suggestion
perl create_jira_tickets.pl [options]
```
I think there's less potential for confusion if we have 'jira' in the script name; it makes it clearer what the script does, and distinguishes it from the existing `create_datacheck.pl` script.

scripts/create_datacheck_tickets.pl
```diff
@@ -0,0 +1,172 @@
...
+
+use Cwd 'abs_path';
+use File::Basename;
+use Getopt::Long;
+use Pod::Usage;
+use POSIX;
+
+use Bio::EnsEMBL::Compara::Utils::JIRA;
+
+use Data::Dumper;
```
james-monkeyshines:
```suggestion
```
Library doesn't seem to be used.

scripts/create_datacheck_tickets.pl
JAlvarezJarreta: This has been addressed already in our `master` branch.

PR #164, opened 2019-10-31
Only load the required DBAdaptor

muffato:
This allows not having every repository (and all their dependencies) set up

### DIFF
--- a/scripts/run_datachecks.pl
+++ b/scripts/run_datachecks.pl
@@ -134,10 +134,6 @@ use strict;
 use feature 'say';
 
 use Bio::EnsEMBL::DataCheck::Manager;
-use Bio::EnsEMBL::DBSQL::DBAdaptor;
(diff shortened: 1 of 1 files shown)


james-monkeyshines:
I'm afraid I don't really like this - I think there needs to be a very good reason to use evals, and I'm not convinced by this use case. Compara, funcgen and variation are all part of the standard set of Ensembl repositories (cf. the 'api' group in [ensembl-git-tools](https://github.com/Ensembl/ensembl-git-tools)), and neither funcgen or variation have had a problem, so I'd like to leave the code as is.
I will add the repos as dependencies in the ensembl-datacheck readme.

muffato:
OK. I made this pull-request because Variation and Funcgen are not part of out standard set of dependencies, so that's extra repos (270 Mb) we need to check out for every branch we're working on (master, e99, dev, etc). We'll have to see how to use symlinks to mitigate the space cost.
The `eval` was exactly the same as used in the Registry.

PR #168, opened 2019-11-12
Feature/compara dcs

CristiGuijarro:
**Further Java HC conversions for compara**

1. NoDataOnGenomeComponents
2. CheckGenomicAlignMTs
3. CheckSyntenySanity
4. CheckTableSizes
5. CompareMSANames

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ before_install:
     - git clone --branch $ENSEMBL_BRANCH --depth 1 https://github.com/Ensembl/ensembl-metadata.git
     - git clone --branch $ENSEMBL_BRANCH --depth 1 https://github.com/Ensembl/ensembl-test.git
     - git clone --branch $ENSEMBL_BRANCH --depth 1 https://github.com/Ensembl/ensembl-variation.git
+    - git clone --branch $ENSEMBL_BRANCH --depth 1 https://github.com/Ensembl/ensembl-compara.git
     - wget https://github.com/bioperl/bioperl-live/archive/release-1-6-924.zip
     - unzip release-1-6-924.zip
 
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
@@ -29,7 +29,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 
 use constant {
   NAME        => 'CheckGenomicAlignMTs',
-  DESCRIPTION => 'There should be an entry in genomic_align per genome with MT cellular components',
+  DESCRIPTION => 'The multiple alignments should include all the MT sequences',
   GROUPS      => ['compara', 'compara_multiple_alignments'],
   DATACHECK_TYPE => 'advisory',
   DB_TYPES    => ['compara'],
@@ -41,9 +41,12 @@ sub tests {
   
   my $helper  = $self->dba->dbc->sql_helper;
   
-  #Collect all the dnafrags that have a mitochondrion component only in the genomic align methods
+  #Collect all the dnafrags that have a mitochondrion component and should be in a multiple alignment
   my $mlss_sql = q/
-    SELECT genome_db.name, method_link_species_set_id, dnafrag_id
+    SELECT genome_db.name AS gdb_name, 
+        method_link_species_set_id, 
+        dnafrag_id, 
+        method_link_species_set.name AS mlss_name
       FROM method_link_species_set 
         JOIN method_link USING(method_link_id)
         JOIN species_set USING(species_set_id)
@@ -57,14 +60,10 @@ sub tests {
   
   my $entries_array = $helper->execute(  
     -SQL => $mlss_sql, 
-    -USE_HASHREFS => 1,
-    -CALLBACK     => sub {
-      my $row = shift @_;
-      return { name => $row->{name}, method_link_species_set_id => $row->{method_link_species_set_id}, dnafrag_id => $row->{dnafrag_id} };
-    },
+    -USE_HASHREFS => 1
   );
   
-  #Checking to makesure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
+  #Checking to make sure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
   foreach my $row (@$entries_array) {
     my $sql = qq/
       SELECT count(*) 
@@ -73,11 +72,10 @@ sub tests {
         AND dnafrag_id = $row->{dnafrag_id}
     /;
     
-    my $desc = "For genome: $row->{name} the mlss_id $row->{method_link_species_set_id} with dnafrag_id $row->{dnafrag_id} is present in genomic_align";
+    my $desc = "The MT of $row->{gdb_name} (dnafrag_id $row->{dnafrag_id}) is not present in the genomic_align table for alignment mlss_id $row->{method_link_species_set_id} ($row->{mlss_name})";
     
     is_rows_nonzero($self->dba, $sql, $desc);
   }
 }
 
 1;
-
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
@@ -23,15 +23,15 @@ use strict;
 
 use Moose;
 use Test::More;
-use Bio::EnsEMBL::Utils::SqlHelper;
 use Bio::EnsEMBL::DataCheck::Test::DataCheck;
 
 extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 
 use constant {
   NAME        => 'CheckGenomicAlignMTs',
-  DESCRIPTION => 'There should be an entry in genomic_align per genome with multiple alignment MLSS',
+  DESCRIPTION => 'There should be an entry in genomic_align per genome with MT cellular components',
   GROUPS      => ['compara', 'compara_multiple_alignments'],
+  DATACHECK_TYPE => 'advisory',
   DB_TYPES    => ['compara'],
   TABLES      => ['dnafrag', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'species_set']
 };
@@ -41,16 +41,17 @@ sub tests {
   
   my $helper  = $self->dba->dbc->sql_helper;
   
+  #Collect all the dnafrags that have a mitochondrion component only in the genomic align methods
   my $mlss_sql = q/
     SELECT genome_db.name, method_link_species_set_id, dnafrag_id
       FROM method_link_species_set 
-        LEFT JOIN method_link USING(method_link_id)
-        LEFT JOIN species_set USING(species_set_id)
-        LEFT JOIN genome_db USING(genome_db_id)
-        LEFT JOIN dnafrag USING(genome_db_id)
+        JOIN method_link USING(method_link_id)
+        JOIN species_set USING(species_set_id)
+        JOIN genome_db USING(genome_db_id)
+        JOIN dnafrag USING(genome_db_id)
       WHERE cellular_component = 'MT' 
         AND (class LIKE 'GenomicAlignTree%' 
-        OR class LIKE 'GenomicAlign%multiple%') 
+          OR class LIKE 'GenomicAlign%multiple%') 
         AND (type NOT LIKE 'CACTUS_HAL%')
     /;
   
@@ -59,19 +60,20 @@ sub tests {
     -USE_HASHREFS => 1,
     -CALLBACK     => sub {
       my $row = shift @_;
-      return { gdb_name => $row->{name}, mlss_id => $row->{method_link_species_set_id}, dnafrag_id => $row->{dnafrag_id} };
+      return { name => $row->{name}, method_link_species_set_id => $row->{method_link_species_set_id}, dnafrag_id => $row->{dnafrag_id} };
     },
   );
   
+  #Checking to makesure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
   foreach my $row (@$entries_array) {
     my $sql = qq/
       SELECT count(*) 
         FROM genomic_align 
-      WHERE method_link_species_set_id = $row->{mlss_id} 
+      WHERE method_link_species_set_id = $row->{method_link_species_set_id} 
         AND dnafrag_id = $row->{dnafrag_id}
     /;
     
-    my $desc = "mlss_id $row->{mlss_id} with dnafrag_id $row->{dnafrag_id} is present in genomic_align";
+    my $desc = "For genome: $row->{name} the mlss_id $row->{method_link_species_set_id} with dnafrag_id $row->{dnafrag_id} is present in genomic_align";
     
     is_rows_nonzero($self->dba, $sql, $desc);
   }
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
@@ -1,0 +1,81 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckGenomicAlignMTs;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::Utils::SqlHelper;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'CheckGenomicAlignMTs',
+  DESCRIPTION => 'There should be an entry in genomic_align per genome with multiple alignment MLSS',
+  GROUPS      => ['compara', 'compara_multiple_alignments'],
+  DB_TYPES    => ['compara'],
+  TABLES      => ['dnafrag', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'species_set']
+};
+
+sub tests {
+  my ($self) = @_;
+  
+  my $helper  = $self->dba->dbc->sql_helper;
+  
+  my $mlss_sql = q/
+    SELECT genome_db.name, method_link_species_set_id, dnafrag_id
+      FROM method_link_species_set 
+        LEFT JOIN method_link USING(method_link_id)
+        LEFT JOIN species_set USING(species_set_id)
+        LEFT JOIN genome_db USING(genome_db_id)
+        LEFT JOIN dnafrag USING(genome_db_id)
+      WHERE cellular_component = 'MT' 
+        AND (class LIKE 'GenomicAlignTree%' 
+        OR class LIKE 'GenomicAlign%multiple%') 
+        AND (type NOT LIKE 'CACTUS_HAL%')
+    /;
+  
+  my $entries_array = $helper->execute(  
+    -SQL => $mlss_sql, 
+    -USE_HASHREFS => 1,
+    -CALLBACK     => sub {
+      my $row = shift @_;
+      return { gdb_name => $row->{name}, mlss_id => $row->{method_link_species_set_id}, dnafrag_id => $row->{dnafrag_id} };
+    },
+  );
+  
+  foreach my $row (@$entries_array) {
+    my $sql = qq/
+      SELECT count(*) 
+        FROM genomic_align 
+      WHERE method_link_species_set_id = $row->{mlss_id} 
+        AND dnafrag_id = $row->{dnafrag_id}
+    /;
+    
+    my $desc = "mlss_id $row->{mlss_id} with dnafrag_id $row->{dnafrag_id} is present in genomic_align";
+    
+    is_rows_nonzero($self->dba, $sql, $desc);
+  }
+}
+
+1;
+
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
@@ -23,9 +23,7 @@ use strict;
 
 use Moose;
 use Test::More;
-use Bio::EnsEMBL::Utils::SqlHelper;
 use Bio::EnsEMBL::DataCheck::Test::DataCheck;
-use Bio::EnsEMBL::Compara::DBSQL::MethodLinkSpeciesSetAdaptor;
 
 extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 
@@ -44,67 +42,53 @@ sub tests {
   my $dba = $self->dba;
   my $helper = $dba->dbc->sql_helper;
   my $mlss_adap = $dba->get_MethodLinkSpeciesSetAdaptor;
+  
   #Collect mlss_ids for synteny methods
-  my $mlss_ids = $mlss_adap->fetch_all_by_method_link_type("SYNTENY");
+  my $mlss = $mlss_adap->fetch_all_by_method_link_type("SYNTENY");
 
-  foreach my $mlss_id ( @$mlss_ids ) {
-    my $mlss = $mlss_id->dbID;
-    #Collect genome_db_ids in mlss_id
-    my $gdb_sql = qq/
-      SELECT genome_db_id 
-        FROM method_link_species_set 
-          LEFT JOIN species_set USING(species_set_id) 
-      WHERE method_link_species_set_id = $mlss
-    /;
-
-    my $gdb_ids = $helper->execute_simple( -SQL => $gdb_sql );
-    #Look for synteny_regions and GenomicAlignBlocks in mlss
-    my $mlss_sql = qq/
-      SELECT mlss2.method_link_species_set_id 
-        FROM method_link_species_set mlss1, method_link_species_set mlss2, method_link ml 
-      WHERE mlss1.method_link_species_set_id = $mlss
-        AND mlss1.species_set_id = mlss2.species_set_id
-        AND mlss2.method_link_id = ml.method_link_id 
-        AND ml.class LIKE 'GenomicAlignBlock%'
-    /;
-
-    my $synteny_mlss = $helper->execute_simple( -SQL => $mlss_sql );
+  foreach my $mlss ( @$mlss ) {
+    my $mlss_id = $mlss->dbID;
+    
+    my $gab_mlss_list = $mlss->get_all_sister_mlss_by_class('GenomicAlignBlock.pairwise_alignment');
+    
+    my $gdb_ids = $mlss->species_set->genome_dbs;
+        
     #Collect dnafrag_ids longer than 1Mb with exceptions
-    foreach my $gdb_id ( @$gdb_ids ) {
+    foreach my $gdb ( @$gdb_ids ) {
+      my $gdb_id = $gdb->dbID;
       my $dnafrag_sql = qq/
       SELECT dnafrag_id 
         FROM dnafrag 
       WHERE genome_db_id = $gdb_id
-        AND coord_system_name IN ('chromosome', 'group')
-        AND name NOT LIKE '%\\_%'
-        AND name NOT LIKE '%Un%'
-        AND name NOT IN ('MT') 
+        AND cellular_component!='MT' 
         AND length > 1000000
       /;
 
-      my $dnafrag_ids = $helper->execute_simple( -SQL => $mlss_sql );
+      my $dnafrag_ids = $helper->execute_simple( -SQL => $dnafrag_sql );
+      
       #Check for reasonable count of synteny regions, fine if more than none
       foreach my $dnafrag_id ( @$dnafrag_ids ) {
         my $synteny_sql = qq/
           SELECT COUNT(*) 
             FROM synteny_region
-              LEFT JOIN dnafrag_region USING (synteny_region_id) 
-          WHERE method_link_species_set_id = $mlss
+              JOIN dnafrag_region USING (synteny_region_id) 
+          WHERE method_link_species_set_id = $mlss_id
             AND dnafrag_id = $dnafrag_id
         /;
+        
         #If no synteny regions counted, check genomic_alignment_blocks
         my $synteny_count = $helper->execute_single_result( -SQL => $synteny_sql );
         my $alignment_count = 0;
         if ( $synteny_count == 0 ) {
-          foreach my $alignment_mlss ( @$synteny_mlss ) {
+          foreach my $gab_mlss ( @$gab_mlss_list ) {
+            #print Dumper($gab_mlss);
+            my $gab_mlss_id = $gab_mlss->method->dbID;
             my $dnafrag_count_sql = qq/
-              SELECT dnafrag.name, count(*) as counter 
+              SELECT ga2.dnafrag_id, count(*) as count 
                 FROM genomic_align ga1 
-                  LEFT JOIN genomic_align ga2 USING (genomic_align_block_id)
-                  LEFT JOIN dnafrag ON (ga2.dnafrag_id = dnafrag.dnafrag_id) 
+                  JOIN genomic_align ga2 USING (genomic_align_block_id)
               WHERE ga1.dnafrag_id = $dnafrag_id
-                AND dnafrag.coord_system_name IN ('chromosome', 'group')
-                AND ga1.method_link_species_set_id = $alignment_mlss
+                AND ga1.method_link_species_set_id = $gab_mlss_id
                 AND ga1.dnafrag_id <> ga2.dnafrag_id 
               GROUP BY ga2.dnafrag_id
               ORDER BY count(*) DESC LIMIT 1
@@ -114,16 +98,19 @@ sub tests {
               -USE_HASHREFS => 1,
               -CALLBACK     => sub {
                 my $row = shift @_;
-                return { dnafrag_name => $row->{name}, count => $row->{counter} };
+                return { dnafrag_id => $row->{dnafrag_id}, count => $row->{count} };
               },
             );
             if ( ( scalar @$aln_array > 0 ) && ( @$aln_array[0]->{count} > $alignment_count ) ) {
               $alignment_count = @$aln_array[0]->{count};
             } 
           }
+
           #Error is reported if there are too many alignments to a dnafrag_id
-          my $desc = "genome_db_id: $gdb_id dnafrag_id.$dnafrag_id has syntenies for MLSS $mlss with <1000 foreign alignments";
+          my $desc = "genome_db_id: $gdb_id dnafrag_id: $dnafrag_id has no syntenies for MLSS: $mlss_id with >1000 inappropriate alignments";
+          
           cmp_ok( $alignment_count, '<', 1000, $desc );
+
         }
       }
     }
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
@@ -38,36 +38,30 @@ use constant {
 
 sub tests {
   my ($self) = @_;
-  
+  # Make sure the registry is loaded
+  $self->registry;
+
   my $dba = $self->dba;
   my $helper = $dba->dbc->sql_helper;
   my $mlss_adap = $dba->get_MethodLinkSpeciesSetAdaptor;
+  my $dnafrag_adap = $dba->get_DnaFragAdaptor;
   
   #Collect mlss_ids for synteny methods
   my $mlss = $mlss_adap->fetch_all_by_method_link_type("SYNTENY");
 
   foreach my $mlss ( @$mlss ) {
     my $mlss_id = $mlss->dbID;
-    
     my $gab_mlss_list = $mlss->get_all_sister_mlss_by_class('GenomicAlignBlock.pairwise_alignment');
+    my $gdbs = $mlss->species_set->genome_dbs;
     
-    my $gdb_ids = $mlss->species_set->genome_dbs;
-        
     #Collect dnafrag_ids longer than 1Mb with exceptions
-    foreach my $gdb ( @$gdb_ids ) {
+    foreach my $gdb ( @$gdbs ) {
       my $gdb_id = $gdb->dbID;
-      my $dnafrag_sql = qq/
-      SELECT dnafrag_id 
-        FROM dnafrag 
-      WHERE genome_db_id = $gdb_id
-        AND cellular_component!='MT' 
-        AND length > 1000000
-      /;
-
-      my $dnafrag_ids = $helper->execute_simple( -SQL => $dnafrag_sql );
+      my $karyo_dnafrags = $dnafrag_adap->fetch_all_karyotype_DnaFrags_by_GenomeDB($gdb);
       
       #Check for reasonable count of synteny regions, fine if more than none
-      foreach my $dnafrag_id ( @$dnafrag_ids ) {
+      foreach my $dnafrag ( @$karyo_dnafrags ) {
+        my $dnafrag_id = $dnafrag->dbID;
         my $synteny_sql = qq/
           SELECT COUNT(*) 
             FROM synteny_region
@@ -78,11 +72,10 @@ sub tests {
         
         #If no synteny regions counted, check genomic_alignment_blocks
         my $synteny_count = $helper->execute_single_result( -SQL => $synteny_sql );
-        my $alignment_count = 0;
         if ( $synteny_count == 0 ) {
+
           foreach my $gab_mlss ( @$gab_mlss_list ) {
-            #print Dumper($gab_mlss);
-            my $gab_mlss_id = $gab_mlss->method->dbID;
+            my $gab_mlss_id = $gab_mlss->dbID;
             my $dnafrag_count_sql = qq/
               SELECT ga2.dnafrag_id, count(*) as count 
                 FROM genomic_align ga1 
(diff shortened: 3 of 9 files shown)


CristiGuijarro:
I am going to fix Travis for the compara repository dependencies

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,81 @@
...
+  
+  my $helper  = $self->dba->dbc->sql_helper;
+  
+  my $mlss_sql = q/
+    SELECT genome_db.name, method_link_species_set_id, dnafrag_id
+      FROM method_link_species_set 
+        LEFT JOIN method_link USING(method_link_id)
+        LEFT JOIN species_set USING(species_set_id)
+        LEFT JOIN genome_db USING(genome_db_id)
+        LEFT JOIN dnafrag USING(genome_db_id)
```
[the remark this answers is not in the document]
CristiGuijarro: Glad you said that, I thought so, but was scared to disagree with the original Java HC query!

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,81 @@
...
+        OR class LIKE 'GenomicAlign%multiple%') 
+        AND (type NOT LIKE 'CACTUS_HAL%')
+    /;
+  
+  my $entries_array = $helper->execute(  
+    -SQL => $mlss_sql, 
+    -USE_HASHREFS => 1,
+    -CALLBACK     => sub {
+      my $row = shift @_;
+      return { gdb_name => $row->{name}, mlss_id => $row->{method_link_species_set_id}, dnafrag_id => $row->{dnafrag_id} };
```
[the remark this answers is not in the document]
CristiGuijarro: Okay, I can use the original names. I was trying to improve readability and reduce text a little

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,81 @@
...
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'CheckGenomicAlignMTs',
+  DESCRIPTION => 'There should be an entry in genomic_align per genome with multiple alignment MLSS',
+  GROUPS      => ['compara', 'compara_multiple_alignments'],
+  DB_TYPES    => ['compara'],
+  TABLES      => ['dnafrag', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'species_set']
+};
```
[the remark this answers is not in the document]
CristiGuijarro: Same reason it has put datacheck_type as 'critical' (your comment below). I think I mistook the way the default datacheck_type is produced by the create_datacheck.pl script. It was intended as advisory, I did not notice

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,130 @@
...
+        #If no synteny regions counted, check genomic_alignment_blocks
+        my $synteny_count = $helper->execute_single_result( -SQL => $synteny_sql );
+        my $alignment_count = 0;
+        if ( $synteny_count == 0 ) {
+          foreach my $alignment_mlss ( @$synteny_mlss ) {
+            my $dnafrag_count_sql = qq/
+              SELECT dnafrag.name, count(*) as counter 
+                FROM genomic_align ga1 
+                  LEFT JOIN genomic_align ga2 USING (genomic_align_block_id)
+                  LEFT JOIN dnafrag ON (ga2.dnafrag_id = dnafrag.dnafrag_id)
```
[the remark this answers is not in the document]
CristiGuijarro: LEFT JOINS will all be replaced, I don't like them either.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,130 @@
...
+    my $gdb_ids = $mlss_id->species_set->genome_dbs;
+        
+    #Collect dnafrag_ids longer than 1Mb with exceptions
+    foreach my $gdb ( @$gdb_ids ) {
+      my $gdb_id = $gdb->dbID;
+      my $dnafrag_sql = qq/
+      SELECT dnafrag_id 
+        FROM dnafrag 
+      WHERE genome_db_id = $gdb_id
+        AND coord_system_name IN ('chromosome', 'group')
```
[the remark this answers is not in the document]
CristiGuijarro: I'll give it a go

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,130 @@
...
+        AND name NOT LIKE '%\\_%'
```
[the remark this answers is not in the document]
CristiGuijarro: I think it is an additional over-the-top filter that is made redundant by the change to cellular_component!='MT', so removed

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,84 @@
...
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'CheckTableSizes',
+  DESCRIPTION    => 'Tables must be populated and not differ significantly in row numbers',
+  GROUPS         => ['compara', 'compara_tables'],
+  DATACHECK_TYPE => 'advisory',
+  DB_TYPES       => ['compara'],
+  TABLES         => []
```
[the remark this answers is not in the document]
CristiGuijarro: I left this empty with the idea that this DC could be used in more ways, for example during production, before all tables have been written and populated, and use on ehive dbs. With this in mind, any issues can be picked up at the end of a pipeline and caught quickly ahead of the end of production just before handover. Do you think this may not be such a good idea? If I put every compara table in the list of TABLES, would the DC still be able to be used during production on pipeline dbs and only partially merged release dbs (for example if we merge just the genomic data and are still waiting on the protein tree data)?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,81 @@
...
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckGenomicAlignMTs;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::Utils::SqlHelper;
```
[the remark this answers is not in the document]
CristiGuijarro: Yes

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,130 @@
...
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckSyntenySanity;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::Utils::SqlHelper;
```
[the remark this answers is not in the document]
CristiGuijarro: Yes, I think so

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,130 @@
...
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use Bio::EnsEMBL::Compara::DBSQL::MethodLinkSpeciesSetAdaptor;
+use Data::Dumper;
+$Data::Dumper::Maxdepth=2;
```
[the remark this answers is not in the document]
CristiGuijarro: Yes, will be removed, I missed it

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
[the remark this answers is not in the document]
muffato: It's probably that at the time the integrity of the database was really poor (foreign keys not enforced ?). Nowadays there are foreign key checks, so all these links _must_ exist

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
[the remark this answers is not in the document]
muffato: As far as I understand `TABLES` is only used in the context of a DC setup that records the status of the tables whenever it is run, and skips tests that touch tables (as declared in `TABLES`) that haven't changed. So we should be able to run the DCs on any database (production, merged, etc) regardless of `TABLES`.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,84 @@
...
+  my $prev_db_name = $prev_dba->dbc->dbname;
+  
+  foreach my $table ( @$curr_tables ) {
+      my $desc_1 = "The number of rows in $table for $curr_db_name has not increased by >10% from $prev_db_name";
+      my $desc_2 = "The number of rows in $table for $curr_db_name has not decreased by >5% from $prev_db_name";
+      my $sql = qq/
+        SELECT COUNT(*) FROM $table
+      /;
+      row_totals($prev_dba, $curr_dba, $sql, $sql, 0.9, $desc_1);
+      row_totals($curr_dba, $prev_dba, $sql, $sql, 0.95, $desc_2);
```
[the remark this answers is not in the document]
muffato: This is missing the test that checks whether the (non-empty) tables have exactly the same number of rows, as it very rarely happens.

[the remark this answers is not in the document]
CristiGuijarro: I will include it

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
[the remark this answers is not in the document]
CristiGuijarro: I take it back, not needed! I feel like it complained once when I didn't have it... But not anymore!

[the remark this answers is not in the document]
muffato: Right. I was going to check :)
Since you are not calling the package directly, you shouldn't need it (and `DBConnection::sql_helper` imports it on its own side)

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,84 @@
...
+  if (defined $reg_conf) {
+    Bio::EnsEMBL::Registry->load_all($reg_conf);
+  }
+  else {
+      die "A compara production_reg_conf needs to be supplied";
+  }
+  
+  my $curr_dba = $self->dba;
+  my $curr_helper = $curr_dba->dbc->sql_helper;
+  my $prev_dba = Bio::EnsEMBL::Compara::DBSQL::DBAdaptor->go_figure_compara_dba('compara_prev');
```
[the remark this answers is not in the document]
CristiGuijarro: `my $prev_dba = $registry->get_DBAdaptor('compara_prev');`

This one doesn't work

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,84 @@
...
+        SELECT COUNT(*) FROM $table
+      /;
+      row_totals($prev_dba, $curr_dba, $sql, $sql, 0.9, $desc_1);
+      row_totals($curr_dba, $prev_dba, $sql, $sql, 0.95, $desc_2);
+  }
+  
+  my $desc = "There are no tables missing in $curr_db_name that were present in $prev_db_name";
+  cmp_ok( scalar(@$curr_tables), '=>', scalar(@$prev_tables), $desc );
+  if ( scalar(@$curr_tables) != scalar(@$prev_tables) ) {
+    diag explain array_diff( $curr_tables, $prev_tables, $curr_db_name, $prev_db_name );
```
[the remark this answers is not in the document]
CristiGuijarro: > `    array_diff
        array_diff($array_1, $array_2, [$array_1_label, [$array_2_label]]);
 
        Calculate the difference between two arrays of string values,
        $array_1 and $array_1, ignoring the order of the elements. This
        method does not work with elements that are references. The
        $array_1_label and $array_2_label parameters allow elements that are
        only in one array to be usefully labelled in the results.`

I was hoping that is what the documentation means. Otherwise, `SHOW TABLES;` does list in alphabetical order by default

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,101 @@
...
+  my $reg_conf = $self->registry_file;
+  if (defined $reg_conf) {
+    Bio::EnsEMBL::Registry->load_all($reg_conf);
+  }
+  else {
+      die "A compara production_reg_conf needs to be supplied";
+  }
+  
+  my $curr_dba = $self->dba;
+  my $prev_dba = Bio::EnsEMBL::Compara::DBSQL::DBAdaptor->go_figure_compara_dba('compara_prev');
```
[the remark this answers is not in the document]
CristiGuijarro: This didn't work, but I shortened the code to make it more visually appealing.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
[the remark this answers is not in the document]
CristiGuijarro: (no)

.travis.yml
```diff
@@ -24,6 +24,7 @@ before_install:
     - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-metadata.git
     - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-test.git
     - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-variation.git
+    - git clone --branch master --depth 1 https://github.com/Ensembl/ensembl-compara.git
```
muffato: This file has actually changed significantly on master since you've opened your pull-request, see https://github.com/Ensembl/ensembl-datacheck/blob/master/.travis.yml#L20-L32
That's why GitHub says there is a conflict on this file. It might be best to rebase the branch ? That would also address the other conflict

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,89 @@
...
+      row_totals( $prev_dba, $curr_dba, $sql, $sql, 0.9, $desc_1 );
+      row_totals( $curr_dba, $prev_dba, $sql, $sql, 0.95, $desc_2 );
+      
+      my $prev_row_count = $prev_helper->execute_single_result( -SQL => $sql );
+
+      cmp_rows( $curr_dba->dbc, $sql, '!=', $prev_row_count, $desc_3 );
+        
+  } else {
+      my $desc_4 = "Table: $table is new in $curr_db_name";
+      pass( $desc_4 );
```
muffato: We want to report all the new tables for manual inspection via a failure. This is to make sure we don't have temporary tables left in the database.
The empty ones should be highlighted because it's unusual to make a schema change to introduce a table but not populate it.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }
+  
+  my $desc = "There are no tables missing in $curr_db_name that were present in $prev_db_name";
+  cmp_ok( scalar(@$curr_tables), '=>', scalar(@$prev_tables), $desc );
+  if ( scalar(@$curr_tables) != scalar(@$prev_tables) ) {
+    diag explain array_diff( $curr_tables, $prev_tables, $curr_db_name, $prev_db_name );
+  }
```
muffato: This won't work if we are adding new tables at the same time. The best is to write a for loop and report the lost tables individually.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,89 @@
...
+      SELECT COUNT(*) FROM $table
+    /;
+      
+    if ( exists($prev_tables{$table}) ) {
+      row_totals( $prev_dba, $curr_dba, $sql, $sql, 0.9, $desc_1 );
+      row_totals( $curr_dba, $prev_dba, $sql, $sql, 0.95, $desc_2 );
+      
+      my $prev_row_count = $prev_helper->execute_single_result( -SQL => $sql );
+
+      cmp_rows( $curr_dba->dbc, $sql, '!=', $prev_row_count, $desc_3 );
```
muffato: Here the SQL query will be run three times on each database. The MySQL server will probably cache the result but it's not guaranteed. I would prefer not using the utility functions `row_totals` and `cmp_rows` and only get the count once from each database. Maybe @james-monkeyshines will have a different opinion...

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,83 @@
...
+        AND (type NOT LIKE 'CACTUS_HAL%')
+    /;
+  
+  my $entries_array = $helper->execute(  
+    -SQL => $mlss_sql, 
+    -USE_HASHREFS => 1,
+    -CALLBACK     => sub {
+      my $row = shift @_;
+      return { name => $row->{name}, method_link_species_set_id => $row->{method_link_species_set_id}, dnafrag_id => $row->{dnafrag_id} };
+    },
```
muffato: Since the columns have the same names as the keys in the hash you can completely skip the `-CALLBACK` parameter. The default hash it returns is what you need

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,83 @@
...
+  #Checking to makesure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
+  foreach my $row (@$entries_array) {
+    my $sql = qq/
+      SELECT count(*) 
+        FROM genomic_align 
+      WHERE method_link_species_set_id = $row->{method_link_species_set_id} 
+        AND dnafrag_id = $row->{dnafrag_id}
+    /;
+    
+    my $desc = "For genome: $row->{name} the mlss_id $row->{method_link_species_set_id} with dnafrag_id $row->{dnafrag_id} is present in genomic_align";
```
muffato:
```suggestion
    my $desc = "The MT of $row->{name} (dnafrag_id $row->{dnafrag_id}) is not present in the genomic_align table for alignment mlss_id $row->{method_link_species_set_id} genomic_align";
```

muffato: What about pulling the mlss name too and adding it to the error message ?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,121 @@
...
+            AND dnafrag_id = $dnafrag_id
+        /;
+        
+        #If no synteny regions counted, check genomic_alignment_blocks
+        my $synteny_count = $helper->execute_single_result( -SQL => $synteny_sql );
+        my $alignment_count = 0;
+        if ( $synteny_count == 0 ) {
+          foreach my $gab_mlss ( @$gab_mlss_list ) {
+            #print Dumper($gab_mlss);
+            my $gab_mlss_id = $gab_mlss->method->dbID;
```
muffato:
```suggestion
            my $gab_mlss_id = $gab_mlss->dbID;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,121 @@
...
+            /;
+            my $aln_array = $helper->execute(  
+              -SQL => $dnafrag_count_sql, 
+              -USE_HASHREFS => 1,
+              -CALLBACK     => sub {
+                my $row = shift @_;
+                return { dnafrag_id => $row->{dnafrag_id}, count => $row->{count} };
+              },
+            );
+            if ( ( scalar @$aln_array > 0 ) && ( @$aln_array[0]->{count} > $alignment_count ) ) {
```
muffato: Since you don't need the dnafrag_id, change the SQL to only return the count (as `COUNT(*)`, no need to give it a name) and use `execute_single_result` instead of `execute`. It will return the count directly

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,121 @@
...
+              $alignment_count = @$aln_array[0]->{count};
+            } 
+          }
+
+          #Error is reported if there are too many alignments to a dnafrag_id
+          my $desc = "genome_db_id: $gdb_id dnafrag_id: $dnafrag_id has no syntenies for MLSS: $mlss_id with >1000 inappropriate alignments";
```
muffato: The alignments are not "inappropriate"
```suggestion
          my $desc = "genome_db_id: $gdb_id dnafrag_id: $dnafrag_id has no syntenies for MLSS: $mlss_id but has >1000 alignments";
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,121 @@
...
+              GROUP BY ga2.dnafrag_id
+              ORDER BY count(*) DESC LIMIT 1
+            /;
+            my $aln_array = $helper->execute(  
+              -SQL => $dnafrag_count_sql, 
+              -USE_HASHREFS => 1,
+              -CALLBACK     => sub {
+                my $row = shift @_;
+                return { dnafrag_id => $row->{dnafrag_id}, count => $row->{count} };
+              },
```
muffato: Could be removed since the column names and the hash key are the same

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,121 @@
...
+            );
+            if ( ( scalar @$aln_array > 0 ) && ( @$aln_array[0]->{count} > $alignment_count ) ) {
+              $alignment_count = @$aln_array[0]->{count};
```
muffato: Do the `cmp_ok` here so that you can add the other dnafrag_id to the error message. 
The algorithm would be simpler: for every alignment mlss_id, count the highest number of of blocks the dnafrag has against another dnafrag and complain if that is more than 1000.
You wouldn't need alignment_count any more

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
muffato: I would be useful to show the mlss_id of the alignment and the other dnafrag_id

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
muffato: This massively changes the output of the script: it does 69599 tests !

muffato:
Don't work on CheckSyntenySanity immediately. I've found a few more things I want to think about first

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
muffato: This makes the test much slower and report a lot of errors :(

muffato:
OK. I think CheckSyntenySanity should be changed to utilize information that was not available in Java. Since syntenies can only be displayed between chromosomes, the test should only focus on those. That was the purpose of the initial filtering on coord_system_name. Now you can use `DnaFragAdaptor::fetch_all_karyotype_DnaFrags_by_GenomeDB` to get the karyotype (chromosomes) of both species and *only* work / report the pairs of dnafrag_ids that are _both_ in the karyotype and have more than 1,000 alignments between them.

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,101 @@
...
+    my $curr_mlss_ids = $curr_mlss_adap->fetch_all_by_method_link_type($mlss_type);
+    my $prev_mlss_ids = $prev_mlss_adap->fetch_all_by_method_link_type($mlss_type);
+    push @curr_mlsses, @$curr_mlss_ids;
+    push @prev_mlsses, @$prev_mlss_ids;
+  }
+  
+  my $curr_count = scalar(@curr_mlsses);
+  my $prev_count = scalar(@prev_mlsses);
+  
+  my $desc_1 = "$curr_db_name has no fewer mlsses than $prev_db_name";
```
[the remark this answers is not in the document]
james-monkeyshines: No particular convention, just need to try and be clear about the intention of the test (which is often quite difficult to do, while remaining succinct). I agree that it's generally easier to understand a positive statement, because if the test fails, you need to negate that to understand the problem. And that's a bit easier if you're not having to juggle double negatives.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,89 @@
...
+};
+
+sub tests {
+  my ($self) = @_;
+  
+  my $curr_dba = $self->dba;
+  my $curr_helper = $curr_dba->dbc->sql_helper;
+  my $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara');
```
muffato: `compara_prev` won't work on the DC service because the latter doesn't use our different registry file. Instead the DC framework provides `$self->get_old_dba`. 
```suggestion
  my $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara') || $self->get_old_dba;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,99 @@
...
+  DATACHECK_TYPE => 'advisory',
+  DB_TYPES       => ['compara'],
+  TABLES         => ['method_link', 'method_link_species_set', 'species_set_header']
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $curr_dba = $self->dba;
+  my $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara');
```
muffato: Same here
```suggestion
  my $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara') || $self->get_old_dba;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,101 @@
...
+  
+  foreach my $mlss ( @prev_mlsses ) {
+    push ( @{ $prev_mlss_names->{$mlss->species_set->name} }, $mlss );
+  }
+  foreach my $mlss ( @curr_mlsses ) {
+    push ( @{ $curr_mlss_names->{$mlss->species_set->name} }, $mlss );
+  }
+  
+  my $desc_2 = "There are no new species_sets in $curr_db_name that were not in $prev_db_name";
+  cmp_ok( ( scalar keys $prev_mlss_names ), '==', ( scalar keys $curr_mlss_names ), $desc_2 );
```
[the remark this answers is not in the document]
CristiGuijarro: Yes, that makes sense. I was considering a lot of these logics more as a warning for the advisory tests. For example, you expect this test to 'fail' if you are expecting more, if the test hasn't picked up an additional MSA of a name but you weer expecting additional, you can see there is an issue. Do you think there is a better way to report this maybe?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
CristiGuijarro: If I understand correctly, I will include another test to check that the new table is populated?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
CristiGuijarro: Yes, this was why I was doing it the way I was before, I thought it made sense to force it to use the registry file we will already have produced. I also thought with providing our own registry, it gives the ability to be more flexible with which previous db is used. I will stick this in anyway.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,82 @@
...
+          OR class LIKE 'GenomicAlign%multiple%') 
+        AND (type NOT LIKE 'CACTUS_HAL%')
+    /;
+  
+  my $entries_array = $helper->execute(  
+    -SQL => $mlss_sql, 
+    -USE_HASHREFS => 1
+  );
+  
+  #Checking to makesure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
```
muffato:
```suggestion
  #Checking to make sure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,82 @@
...
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'CheckGenomicAlignMTs',
+  DESCRIPTION => 'There should be an entry in genomic_align per genome with MT cellular components',
```
muffato:
```suggestion
  DESCRIPTION => 'The multiple alignments should include all the MT sequences',
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,82 @@
...
+  DB_TYPES    => ['compara'],
+  TABLES      => ['dnafrag', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'species_set']
+};
+
+sub tests {
+  my ($self) = @_;
+  
+  my $helper  = $self->dba->dbc->sql_helper;
+  
+  #Collect all the dnafrags that have a mitochondrion component only in the genomic align methods
```
muffato:
```suggestion
  #Collect all the dnafrags that have a mitochondrion component and should be in a multiple alignment
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -0,0 +1,82 @@
...
+  #Checking to makesure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
+  foreach my $row (@$entries_array) {
+    my $sql = qq/
+      SELECT count(*) 
+        FROM genomic_align 
+      WHERE method_link_species_set_id = $row->{method_link_species_set_id} 
+        AND dnafrag_id = $row->{dnafrag_id}
+    /;
+    
+    my $desc = "The MT of $row->{gdb_name} (dnafrag_id $row->{dnafrag_id}) is not present in the genomic_align table for alignment mlss_id $row->{method_link_species_set_id} ($row->{mlss_name}) genomic_align";
```
muffato:
```suggestion
    my $desc = "The MT of $row->{gdb_name} (dnafrag_id $row->{dnafrag_id}) is not present in the genomic_align table for alignment mlss_id $row->{method_link_species_set_id} ($row->{mlss_name})";
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,105 @@
...
+  DESCRIPTION    => 'Check for missing syntenies in the compara database',
+  GROUPS         => ['compara', 'compara_syntenies'],
+  DATACHECK_TYPE => 'advisory',
+  DB_TYPES       => ['compara'],
+  TABLES         => ['dnafrag', 'dnafrag_region', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'synteny_region']
+};
+
+sub tests {
+  my ($self) = @_;
+  $self->registry->load_all();
```
muffato: OK, I understand what you meant. Within the DCs the Registry is only loaded when `$self->registry` is invoked. Here you don't need to call `load_all()` on it
```suggestion
  # Make sure the registry is loaded
  $self->registry

```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,90 @@
...
+  my $curr_db_name = $curr_dba->dbc->dbname;
+  my $prev_db_name = $prev_dba->dbc->dbname;
+  my $curr_mlss_adap = $curr_dba->get_MethodLinkSpeciesSetAdaptor;
+  my $prev_mlss_adap = $prev_dba->get_MethodLinkSpeciesSetAdaptor;
+  my @mlss_types = qw ( PECAN EPO EPO_LOW_COVERAGE CACTUS_HAL );
+  my @curr_mlsses;
+  my @prev_mlsses;
+  
+  foreach my $mlss_type ( @mlss_types ) {
+    my $curr_mlss_ids = $curr_mlss_adap->fetch_all_by_method_link_type($mlss_type);
```
muffato:
```suggestion
    my $curr_mlsses = $curr_mlss_adap->fetch_all_by_method_link_type($mlss_type);
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,90 @@
...
+    my $prev_mlss_ids = $prev_mlss_adap->fetch_all_by_method_link_type($mlss_type);
```
muffato:
```suggestion
    my $prev_mlsses = $prev_mlss_adap->fetch_all_by_method_link_type($mlss_type);
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,90 @@
...
+    push @curr_mlsses, @$curr_mlss_ids;
```
muffato:
```suggestion
    push @curr_mlsses, @$curr_mlsses;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,90 @@
...
+    push @prev_mlsses, @$prev_mlss_ids;
```
muffato:
```suggestion
    push @prev_mlsses, @$prev_mlsses;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -0,0 +1,90 @@
...
+  
+  my $desc_2 = "The number of species_sets in $curr_db_name are the same as in $prev_db_name";
+  cmp_ok( ( scalar keys $curr_mlss_names ), '==', ( scalar keys $prev_mlss_names ), $desc_2 );
+  
+  foreach my $species_set_name ( keys $prev_mlss_names ) {
+    my $desc_4 = "$species_set_name is present in $curr_db_name";
+    if ( exists ($curr_mlss_names->{$species_set_name}) ) {
+      pass( $desc_4 );
+      my $desc_3 = "The number of multiple alignments for species_set <$species_set_name> in $curr_db_name is the same or more than in $prev_db_name";
+      cmp_ok( @{ $curr_mlss_names->{$species_set_name} }, '=>', @{ $prev_mlss_names->{$species_set_name} }, $desc_3 );
```
muffato:
```suggestion
      cmp_ok( @{ $curr_mlss_names->{$species_set_name} }, '<=', @{ $prev_mlss_names->{$species_set_name} }, $desc_3 );
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,106 @@
...
+              ORDER BY count(*) DESC LIMIT 1
+            /;
+            my $aln_array = $helper->execute(  
+              -SQL => $dnafrag_count_sql, 
+              -USE_HASHREFS => 1,
+            );
+            
+            next if ( scalar @$aln_array == 0 );
+            
+            my $desc_1 = "genome_db_id: $gdb_id (" . $gdb->name . ") dnafrag_id: $dnafrag_id has no syntenies for mlss: $mlss_id but has >1000 alignments in the genomic_align_block mlss: $gab_mlss_id with dnafrag_id: " . @$aln_array[0]->{dnafrag_id};
```
muffato:
```suggestion
            my $desc_1 = "genome_db_id: $gdb_id (" . $gdb->name . ") dnafrag_id: $dnafrag_id has no syntenies for mlss: $mlss_id but has <1000 alignments in the genomic_align_block mlss: $gab_mlss_id with dnafrag_id: " . @$aln_array[0]->{dnafrag_id};
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,106 @@
...
+  GROUPS         => ['compara', 'compara_syntenies'],
+  DATACHECK_TYPE => 'advisory',
+  DB_TYPES       => ['compara'],
+  TABLES         => ['dnafrag', 'dnafrag_region', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'synteny_region']
+};
+
+sub tests {
+  my ($self) = @_;
+  # Make sure the registry is loaded
+  $self->registry
```
CristiGuijarro:
```suggestion
  $self->registry;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,95 @@
...
+      pass( $desc_4 );
+      my $desc_5 = "New table: $table is populated with data";
+      is_rows_nonzero($curr_dba->dbc, $sql, $desc_5);
+    }
+  }
+  
+  foreach my $table ( @$prev_tables ) {
+    my $desc_5 = "Table $table is missing in $curr_db_name and present in $prev_db_name";
+    unless ( exists($curr_tables{$table}) ) {
+      fail($desc_5);
```
james-monkeyshines: Generally speaking, it's best to avoid explicit calls to `fail` (and `pass`) (cf Perl's Test::More docs, at the bottom of this section: https://perldoc.perl.org/Test/More.html#I'm-ok%2c-you're-not-ok.)
Here, you could just have the `exists($curr_tables{$table})` in an `ok` statement, and remove the conditional loop and the need for `fail`.
(Up above, it seems like the `pass` isn't really necessary, the test on line 81 implicitly tells you that. It's also a bit confusing to use `$desc_5` for two different descriptions.)

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,95 @@
...
+    my $desc_3 = "The number of rows in $table has not remained static between $curr_db_name and $prev_db_name";
+    my $sql = qq/
+      SELECT COUNT(*) FROM $table
+    /;
+      
+    if ( exists($prev_tables{$table}) ) {
+      my $prev_row_count = $prev_helper->execute_single_result( -SQL => $sql );
+      my $curr_row_count = $curr_helper->execute_single_result( -SQL => $sql );
+
+      cmp_ok( $curr_row_count, '<=', ($prev_row_count*1.1), $desc_1 );
```
james-monkeyshines: Note that there is a method in Bio::EnsEMBL::DataCheck::Test::DataCheck, `row_totals` which wraps this functionality up and enables a bit more succinctness. Admittedly, you would be doing the count multiple times, rather than using the count variable - but we can probably assume that mysql caching will mitigate that. `row_totals` is also perhaps a bit less readable, so I can see why you wouldn't want to use it, but just wanted to make sure you knew it was there...

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
```diff
@@ -0,0 +1,106 @@
...
+  GROUPS         => ['compara', 'compara_syntenies'],
+  DATACHECK_TYPE => 'advisory',
+  DB_TYPES       => ['compara'],
+  TABLES         => ['dnafrag', 'dnafrag_region', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'synteny_region']
+};
+
+sub tests {
+  my ($self) = @_;
+  # Make sure the registry is loaded
+  $self->registry;
```
james-monkeyshines: I wouldn't have thought that this would be necessary - I realise it was a while ago now, but if you could remember why you needed it, that would be helpful, as that might be something I need to take a look at.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -0,0 +1,95 @@
...
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckTableSizes;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use Bio::EnsEMBL::DataCheck::Utils qw/ array_diff /;
```
james-monkeyshines: Minor point, seems like this isn't used in the module?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
CristiGuijarro: I will fix some of these stylistic changes before you merge, I prefer your idea here

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
CristiGuijarro: Thank you, I did use `row_totals` in another version of this CheckTableSizes, but decided on this way for the reasons you have mentioned. I think I have used it in another DC however.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckSyntenySanity.pm
CristiGuijarro: It is not necessary, I think that is a line I forgot to remove when I was running some early tests, I will take it away!

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
CristiGuijarro: It used to be used, another one I have forgotten to remove

PR #175, opened 2019-12-10
Fix to ensure same sorting in SQL results and Perl array.

james-monkeyshines:


james-monkeyshines:
It's underscores that are the problem. Using the default collation, the SQL sort will put underscores after letters; perl will sort underscores first.

Anyway, I'll add an update to the test db to ensure that Travis is happy, the failure is due to a schema change.

marcoooo:
> It's underscores that are the problem. Using the default collation, the SQL sort will put underscores after letters; perl will sort underscores first.
> 
> Anyway, I'll add an update to the test db to ensure that Travis is happy, the failure is due to a schema change.

Thanks for the explanation James, didn't know Perl / SQL ordering difference (dumbly thought that they both use the same). Great then.

james-monkeyshines:
> > It's underscores that are the problem. Using the default collation, the SQL sort will put underscores after letters; perl will sort underscores first.
> > Anyway, I'll add an update to the test db to ensure that Travis is happy, the failure is due to a schema change.
> 
> Thanks for the explanation James, didn't know Perl / SQL ordering difference (dumbly thought that they both use the same). Great then.

I didn't know either, until Bruno pointed out that the datacheck wasn't working...

PR #179, opened 2019-12-20
Some vertebrate species need more memory

thomasmaurel:


james-monkeyshines:
Obviously this is required for now, but it would be better to have a low default (since most jobs require a lot less memory), and have the pipeline fall through to a high-mem version when necessary. I've created a jira ticket for that.

PR #182, opened 2020-01-08
Reduce drosohpila test db short cut to something sustainable with lon…

james-monkeyshines:
…g user names :-) Feel free to refuse it if you think this is not worth it (I can't change my user name)


james-monkeyshines:
I created a pull request so I could say why I'm refusing it :)
The tests fail with the change to 'drosophila' because the database adaptor is created using that string, and so the 'drosophila' species is not found. I had a quick poke around for obvious workarounds, but couldn't find anything...

PR #187, opened 2020-01-29
Updates to support GRCh37 and Pan

muffato:
Changes required to make the DCs pass on the GRCh37 and Pan databases

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
@@ -81,21 +81,22 @@ sub tests {
     -USE_HASHREFS => 1
   );
   
-  #Checking to make sure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
-  foreach my $row (@$entries_array) {
-    my $sql = qq/
-      SELECT count(*) 
-        FROM genomic_align 
-      WHERE method_link_species_set_id = $row->{method_link_species_set_id} 
-        AND dnafrag_id = $row->{dnafrag_id}
-    /;
-    
-    my $desc = "The MT of $row->{gdb_name} (dnafrag_id $row->{dnafrag_id}) is not present in the genomic_align table for alignment mlss_id $row->{method_link_species_set_id} ($row->{mlss_name})";
-    
-    is_rows_nonzero($self->dba, $sql, $desc);
-  }
-  unless (@$entries_array) {
-    pass('None of the species included in multiple alignments have a mitochondrion');
+  SKIP: {
+    skip 'None of the species included in multiple alignments have a mitochondrion' unless scalar(@$entries_array);
+
+    #Checking to make sure that there is at least one row in genomic for each mlss with dnafrag_id linked to MT
+    foreach my $row (@$entries_array) {
+      my $sql = qq/
+        SELECT count(*)
+          FROM genomic_align
+        WHERE method_link_species_set_id = $row->{method_link_species_set_id}
+          AND dnafrag_id = $row->{dnafrag_id}
+      /;
+
+      my $desc = "The MT of $row->{gdb_name} (dnafrag_id $row->{dnafrag_id}) is not present in the genomic_align table for alignment mlss_id $row->{method_link_species_set_id} ($row->{mlss_name})";
+
+      is_rows_nonzero($self->dba, $sql, $desc);
+    }
   }
 }
 
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
@@ -36,6 +36,24 @@ use constant {
   TABLES      => ['dnafrag', 'genome_db', 'genomic_align', 'method_link', 'method_link_species_set', 'species_set']
 };
 
+sub skip_tests {
+    my ($self) = @_;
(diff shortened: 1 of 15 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CheckOrthologyQCThresholds.pm
```diff
@@ -39,18 +40,34 @@ use constant {
...
+    SELECT COUNT(*)
+      FROM method_link_species_set_attr
+    WHERE wga_quality_threshold IS NOT NULL
+  /;
+  my $has_wga_scores = sql_count($self->dba, 'SELECT 1 FROM homology WHERE wga_coverage IS NOT NULL LIMIT 1');
+  if ($has_wga_scores) {
+    my $desc = "There are some wga_quality_threshold in method_link_species_set_attr";
+    is_rows_nonzero( $dbc, $wga_sql, $desc );
+  } else {
+    my $desc = "There are no wga_quality_threshold in method_link_species_set_attr";
```
CristiGuijarro: To stick with consistency, the test descriptions should be passing descriptions only. There shouldn't be a need to include an else with the opposing statement, @james-monkeyshines mentioned in one of my PRs that you can include a SKIP block, perhaps this here would be a good place to implement it?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignMTs.pm
```diff
@@ -76,6 +94,9 @@ sub tests {
     
     is_rows_nonzero($self->dba, $sql, $desc);
   }
+  unless (@$entries_array) {
```
CristiGuijarro: A SKIP BLOCK here?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckOrthologyQCThresholds.pm
muffato: The test in the _else_ block is actually different. On one side it tests that there are rows, on the other side it tests that there no rows. I could combine both tests into something like:
- get the number of rows
- depending on the `has*` boolean, set the comparison operator to `==` or `>`
- use `cmp_ok` with a description like "There is the expected number of rows with wga_quality_threshold"

CristiGuijarro: I like that idea with the comparison operator, that way it is just a single positive description for both

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
```diff
@@ -40,7 +40,10 @@ sub tests {
   
   my $curr_dba = $self->dba;
   my $curr_helper = $curr_dba->dbc->sql_helper;
-  my $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara') || $self->get_old_dba;
+  my $prev_dba;
+  $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara') if $self->registry_file;
+  $prev_dba = $self->get_old_dba if $self->old_server_uri && !$prev_dba;
+  die "Neither -registry_file nor -old_server_uri given on the command-line. Cannot find the previous database" unless $prev_dba;
```
james-monkeyshines: Generally speaking, it's better to throw rather than die in a hive module. (But are you sure you want to die here? If this datacheck is being run as part of a group of datachecks, that will abort all of them. I know I've previously said it's best to avoid using `fail` directly, but rules are there to be broken...)
And since this could be run within a pipeline, might be better to remove reference to the command line, to avoid confusion.
```suggestion
  $self->throw("Neither 'registry_file' nor 'old_server_uri' parameter given. Cannot find the previous database") unless $prev_dba;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareMSANames.pm
```diff
@@ -38,7 +38,10 @@ sub tests {
   my ($self) = @_;
 
   my $curr_dba = $self->dba;
-  my $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara') || $self->get_old_dba;
+  my $prev_dba;
+  $prev_dba = $self->registry->get_DBAdaptor('compara_prev', 'compara') if $self->registry_file;
+  $prev_dba = $self->get_old_dba if $self->old_server_uri && !$prev_dba;
+  die "Neither -registry_file nor -old_server_uri given on the command-line. Cannot find the previous database" unless $prev_dba;
```
james-monkeyshines: See comments on CheckTableSizes.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
muffato: Right, good point. I think I would prefer using `fail` and doing an early `return` instead.

muffato:
Thanks !

lib/Bio/EnsEMBL/DataCheck/Checks/CheckTableSizes.pm
muffato: By the way, there are dozens of `die` statements in the repo (BaseCheck.pm, DbCheck.pm, Checks/*.pm)

james-monkeyshines: Ah, yes... It's OK in BaseCheck and DbCheck, because those are part of the infrastructure; if something is wrong there, then you probably do want everything to fall down around your ears. I'll create a ticket to look into the Checks/*.pm though. Thanks!

PR #188, opened 2020-02-06
Bug correction. Test of proportion of fragment with distance to paren…

dthybert:
the datacheck is checking that the number of entries with a distance to parent >1 is less than 1% of the total fragment. 
The calculation of the ratio was done the way around so to corect the two requests have been inverted to calculate the right ratio.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignTreeTable.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckGenomicAlignTreeTable.pm
@@ -98,7 +98,7 @@ sub tests {
     
     my $desc_3 = "Less than 1% of entries have a distance_to_parent > 1 for $mlss_id ($mlss_name)";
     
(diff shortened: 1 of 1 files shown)


muffato:
I would also like to note that the semantics are not right at two levels. The description of `row_totals` is:
> For example, a value of 0.75 means that the count for $dbc1>must not be less than 75% of the count for $dbc2

It tests that the count in a certain query is at least a certain value, obtained from a reference query

First of all, in `row_totals` itself: to fit the spirit of `cmp_ok` spirit, i.e. `cmp_ok( $got, $op, $expected, $test_name )`, I think it should be implemented as
```
cmp_ok( $count1, '>=', $count2 * $min_proportion, $name )
```

Then, I don't think we should be using it here because it checks that the count is _above_ a minimum value, whereas we want to check that the count is _below_ a maximum value.
Of course `a <= b` is the same as `b >= a`, but it matters
1) when the test fails because the diagnostic message is inverted, and we don't understand what our count is and what the reference count is
2) when setting `min_proportion` (here we need to set it to 100 to achieve 1% ??)

PR #192, opened 2020-02-12
Bugfix: check gene_tree count, not family count

JAlvarezJarreta:


muffato:
Agreed. Already proposed in #187. https://github.com/Ensembl/ensembl-datacheck/pull/187/files#diff-da3de5808f910ea2b9762d5951cff83a

PR #196, opened 2020-02-20
Feature/compara dc

CristiGuijarro:
Two new compara datachecks to ensure that the number of rows between the previous and current release have not dropped for wga_coverage and goc_score following issue in release/100 for WGA scores.

I have put these into two different checks because they are generated by two different pipelines.

There is already a datacheck: `CheckOrthologyQCThresholds.pm` that tests there is at least one row for each of wga_coverage and goc_score that IS NOT NULL. Possibly this could be swapped out, or these two new datachecks merged into it?

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/index.json
+++ b/lib/Bio/EnsEMBL/DataCheck/index.json
@@ -258,6 +258,16 @@
       "name" : "CheckFlatProteinTrees",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::CheckFlatProteinTrees"
    },
+   "CheckGOCScoreStats" : {
+      "datacheck_type" : "critical",
+      "description" : "The number of rows for GOC have not dropped from previous release",
+      "groups" : [
+         "compara",
+         "compara_protein_trees"
+      ],
+      "name" : "CheckGOCScoreStats",
+      "package_name" : "Bio::EnsEMBL::DataCheck::Checks::CheckGOCScoreStats"
+   },
    "CheckGeneGainLossData" : {
       "datacheck_type" : "critical",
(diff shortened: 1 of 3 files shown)


muffato:
I think you can leave `CheckOrthologyQCThresholds` as it is. It uses this at-least-one-row check to configure itself, to know what to expect (the same way other DCs check the presence of MLSSs of certain types)

lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -258,6 +258,18 @@
       "name" : "CheckFlatProteinTrees",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::CheckFlatProteinTrees"
    },
+   "CheckGOCScoreStats" : {
+      "datacheck_type" : "critical",
+      "description" : "The number of rows for GOC have not dropped from previous release",
+      "groups" : [
+         "compara",
+         "compara_multiple_alignments",
+         "compara_pairwise_alignments",
```
muffato: These two checks don't apply to alignment databases

lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -477,6 +489,18 @@
       "name" : "CheckTableSizes",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::CheckTableSizes"
    },
+   "CheckWGACoverageStats" : {
+      "datacheck_type" : "critical",
+      "description" : "The number of rows for WGA coverage have not dropped from previous release",
+      "groups" : [
+         "compara",
+         "compara_multiple_alignments",
+         "compara_pairwise_alignments",
```
muffato: Same here

PR #197, opened 2020-02-20
fk() exception for the ancestral_sequences in genome_db

CristiGuijarro:
A constraint has been added to exclude the `genome_db.name` `ancestral_sequences` as this is not expected to be in any `species_sets`.


muffato:
Duplicate of #190

PR #204, opened 2020-03-09
Adding a new datacheck to make sure new species production name when …

thomasmaurel:
…converted into mart dataset and processed with martbuilder name doesn't exceed the MySQL maximum char limit of 64. Also checks that a species mart dataset name is unique across all the divisions using the metadata database.
The downside of this check is that it create new dependencies on the ensembl-biomart and ensembl-metadata repos.


thomasmaurel:
I believe this PR will fix Travis as some code on branch 100 has not been merged to master yet: https://github.com/Ensembl/ensembl-biomart/pull/109

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+      }
+    }
+    # If dataset is longer than 18 char, we won't be able to generate gene mart tables
+    # like dataset_gene_ensembl__protein_feature_superfamily__dm as this will exceed
+    # the MySQL table name limit of 64 char.
+    if (length($mart_dataset) > 18) {
+      fail("$mart_dataset name is too long. Check with data team and remove this species $dbname from the include ini file\n");
+      return;
+    }
+    my $metadata_dba = $self->registry->get_DBAdaptor('multi', 'metadata');
```
james-monkeyshines: It's best to use the DbCheck `get_dba` method for this; as well as looking a bit neater, that ensures that the db connection will get properly cleared up when the datacheck completes.
```suggestion
    my $metadata_dba = $self->get_dba('multi', 'metadata');
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+    my $gdba = $metadata_dba->get_GenomeInfoAdaptor();
+    my $rdba = $metadata_dba->get_DataReleaseInfoAdaptor();
+    my ($release,$release_info,$included_species);
```
james-monkeyshines: None of these variables are used anywhere...

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+        my $other_species_mart_dataset_name = generate_dataset_name_from_db_name($genome->dbname);
+        # If the species mart databaset clash with another species report it.
+        if($mart_dataset eq $other_species_mart_dataset_name){
+          fail("Mart dataset name $mart_dataset for $production_name ($dbname) clashes with ".$genome->name." (".$genome->dbname."). Please change production and database name");
+          return;
+        }
+      }
+    }
+  }
+  ok('Biomart tests all successful');
```
james-monkeyshines: Generally, `ok`s are to be avoided in the same way as `fail`s - if you avoid those earlier `fail`s, this is unnecessary.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+    # Get list of divisions
+    my $divisions = $gdba->list_divisions();
+    foreach my $div (@$divisions){
+      # Exlude Bacteria since we don't have a mart for them and they slow down this test
+      next if $div eq "EnsemblBacteria";
+      # Get all the genomes for the current release and division
+      my $genomes = $gdba->fetch_all_by_division($div);
+      foreach my $genome (@$genomes){
+        # Skip this species
+        if ($genome->name eq $production_name){
```
james-monkeyshines: If the skip block above is moved as I suggest, we won't have the $production_name variable, but can just use $self->species instead.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckMartDatasetName;
```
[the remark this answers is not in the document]
thomasmaurel: Thanks @james-monkeyshines. That's what happen when you copy someone else DC (which I should not be doing)

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use MartUtils qw(generate_dataset_name_from_db_name);
+use Bio::EnsEMBL::BioMart::Mart qw(genome_to_include);
```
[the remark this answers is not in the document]
thomasmaurel: @james-monkeyshines: Yes, I did that reluctantly. My worries is that if we change the mart regex we need to change the DC too. Thinking about it, it makes no sense to force all the teams to have a checkout of the biomart and metadata repos for one test only.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+  DATACHECK_TYPE => 'critical',
+  DB_TYPES       => ['core'],
+  TABLES         => ['meta']
+};
+
+sub tests {
+  my ($self) = @_;
+  SKIP: {
+    my $dbname = $self->dba->dbc->dbname;
+    my $mart_dataset = generate_dataset_name_from_db_name($dbname);
```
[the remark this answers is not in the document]
thomasmaurel: Very true, I've moved the code into skip_tests, you are right it's cleaner

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+      my $included_species = genome_to_include($division,$repo_location);
+      if (!grep( /$production_name/, @$included_species) ){
+        skip 'We dont build mart for this species', 1;
+      }
+    }
+    # If dataset is longer than 18 char, we won't be able to generate gene mart tables
+    # like dataset_gene_ensembl__protein_feature_superfamily__dm as this will exceed
+    # the MySQL table name limit of 64 char.
+    if (length($mart_dataset) > 18) {
+      fail("$mart_dataset name is too long. Check with data team and remove this species $dbname from the include ini file\n");
```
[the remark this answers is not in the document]
thomasmaurel: Yes, make sense. Again, I copied code from somewhere else. I also improved the diagnostic message, hope it makes more sense now.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
thomasmaurel: Sure, thanks for letting me know about this.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
```diff
@@ -0,0 +1,100 @@
...
+      foreach my $genome (@$genomes){
+        # Skip this species
+        if ($genome->name eq $production_name){
+          next;
+        }
+        # Generate mart name using regexes
+        my $other_species_mart_dataset_name = generate_dataset_name_from_db_name($genome->dbname);
+        # If the species mart databaset clash with another species report it.
+        if($mart_dataset eq $other_species_mart_dataset_name){
+          fail("Mart dataset name $mart_dataset for $production_name ($dbname) clashes with ".$genome->name." (".$genome->dbname."). Please change production and database name");
```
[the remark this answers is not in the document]
thomasmaurel: Yes, pushing to an array is a great idea. Thanks.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
thomasmaurel: Thanks, I forgot about $self->species. It's nice to use.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
thomasmaurel: Yes, sorry, again copy/pasting some code 🙄

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
thomasmaurel: Sure, I've removed that too thanks.

thomasmaurel:
@james-monkeyshines: I made a couple of changes since your review:
1) I am skipping collections databases since these are not in mart
2) I am only checking vertebrates against vertebrates for duplicates and non-vert against non-vert (excluding Bacteria). I've spotted some conflicts between vert and non-vert names. Since these are on different server for now, I think we can ignore these
3) I changed the code loading the list of species for vertebrates from a file to using the GitHub repo. This remove the needs for teams to have the biomart repo checked out and on the correct branch. 
4) I slightly changed the char limit for mart name, actually a species can be 18 charfusarium_pseudograminearum just hit the 64 char limit of MySQL but don't go over.
5) I ran this DC on all the vert and non-vert core databases to make sure it would not prevent handover of current database. The only thing I have spotted is the clash between Apple and Fly.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMartDatasetName.pm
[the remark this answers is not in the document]
thomasmaurel: I made extra changes to load the list of species for vert from GitHub to remove dependancy on the ensembl-biomart repo. I can also add the DC to this page so that if we update the regex, we will remember to update this as well: https://www.ebi.ac.uk/seqdb/confluence/display/GTI/Change+regex+used+to+generate+mart+internal+names

PR #206, opened 2020-03-10
Misc. fixes

muffato:
Hello

These are four unrelated changes:
1. Minor comment fix
2. Test the objects as they come instead of preloading them into memory (don't need to allocate GBs of RAM any more to run the DCs)
3. Fix a Registry bug whereby database connections are closed at the handle level without letting the object now. Causes "MySQL server has gone away" errors
4. Got rid of some Compara-specific code


james-monkeyshines:
I'll merge following a review from a compara team member.

PR #208, opened 2020-03-17
Ensprod 4980

vinay-ebi:
Converted DescriptionNewline Health check to datacheck

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionNewlines.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionNewlines.pm
@@ -30,22 +30,24 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME           => 'DescriptionNewlines',
   DESCRIPTION    => 'Check for newlines and tabs in gene descriptions',
-  GROUPS         => ['xref'],
+  GROUPS         => ['xref', 'core'],
   DATACHECK_TYPE => 'critical',
   TABLES         => ['gene']
 };
 
 sub tests {
-  my ($self) = @_;
 
+  my ($self) = @_;
+  my $species_id = $self->dba->species_id;
   my $desc_1 = 'gene description does not contain  newlines and/or tabs';
-  my $sql_1  = qq/
-  	SELECT COUNT(*) 
-        FROM gene 
-        WHERE (LOCATE('\n', description) > 0 OR LOCATE('\t', description) > 0)
+  my $sql_2 = qq/
+    SELECT count(*) FROM  gene g 
+    INNER JOIN seq_region sr USING (seq_region_id) 
+    INNER JOIN  coord_system cs USING (coord_system_id)   
+    where cs.species_id = $species_id  AND (LOCATE('\n', g.description) > 0 OR LOCATE('\t', g.description) > 0)
   /;
 
-  is_rows_zero($self->dba, $sql_1, $desc_1);
+  is_rows_zero($self->dba, $sql_2, $desc_1);
 }
 
 1;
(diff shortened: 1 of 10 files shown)


vinay-ebi:
This PR include code for two tickets : ENSPROD-4980 and ENSPROD-4748

lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionNewlines.pm
```diff
@@ -0,0 +1,52 @@
...
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'DescriptionNewlines',
+  DESCRIPTION    => 'Check for newlines and tabs in gene descriptions',
+  GROUPS         => ['xref'],
+  DATACHECK_TYPE => 'critical',
```
james-monkeyshines: It doesn't hurt to have this here, but this is the default type, so it's not obligatory.

lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionNewlines.pm
```diff
@@ -0,0 +1,52 @@
...
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'DescriptionNewlines',
+  DESCRIPTION    => 'Check for newlines and tabs in gene descriptions',
+  GROUPS         => ['xref'],
```
james-monkeyshines:
```suggestion
  GROUPS         => ['xref', 'core'],
```
All of the new set of xref-related datachecks need to be in the 'core' group, which is run when dbs are handed over - because non-vert dbs will already have xrefs that we need to check (an alternative approach would be to run multiple groups at handover, but I think people like to have a single group that they can run, and know that it will be fine for handover if everything passes).

lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionNewlines.pm
```diff
@@ -0,0 +1,52 @@
...
+  TABLES         => ['gene']
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $desc_1 = 'gene description does not contain  newlines and/or tabs';
+  my $sql_1  = qq/
+  	SELECT COUNT(*) 
+        FROM gene
```
james-monkeyshines: Collection databases necessitate some joins here. The issue is that this test will be run once per species - for standard core dbs that's fine, because 1 db = 1 species. A collection db can have up to 250 species; running the same query 250 times is wasteful, and if there is a problem, it's not immediately obvious which species are causing the problem.

The solution is to make the query species-specific, for which we need to join to the `coord_system` table, via `seq_region`, in order to apply the necessary condition. For example, https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/UniProtDisplayXref.pm#L48.

It's not applicable here, but some tables are not species-specific, in which case you can set the `-per_db` flag when creating the datacheck - this will ensure that for collection dbs, it only gets run once. E.g. https://github.com/Ensembl/ensembl-datacheck/blob/master/lib/Bio/EnsEMBL/DataCheck/Checks/DuplicateXref.pm#L36

If you're wondering why the originating healthcheck didn't have the correct SQL, that's because healthchecks can sometimes pre-date collection dbs; in other cases they were just not written with collection dbs in mind.

PR #214, opened 2020-03-20
Misc Compara changes

muffato:
This PR was intended to replace the `class` filters with `type` in several places in order to be more consistent across the HCs, but I also found two occurrences of `compara_prev` that needed to be replaced (they were introduced in parallel of my PR #206 , that's why I had missed them).

I've left some occurrences of `class` in ForeignKeysCompara because it's meant to be really generic, but all others DCs are now using `type` only.

(leaving this as a draft because I want to check something with Cristi first)


muffato:
Added two more commits to avoid "Use of uninitialized value" warnings

PR #216, opened 2020-03-20
Datachecks for xref

vinay-ebi:
Data checks for xref table:
 1) Tests for {ECO: } blocks from Uniprot in descriptions.
 2) Test for Bad characters in display_label
 3) Tests for improper import of MIM data.


lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionDisplayXrefs.pm
```diff
@@ -0,0 +1,77 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::DescriptionDisplayXrefs;
```
james-monkeyshines: I think a better name might be 'XrefFormat' - provides a succinct summary of what we're testing.

james-monkeyshines: Having suggested that, it sounded like it rang a bell, and there is already a datacheck of that name, covering the same ground. So the tests here should be added to that, rather than creating a new one.

lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionDisplayXrefs.pm
```diff
@@ -0,0 +1,77 @@
...
+      SELECT count(*) FROM xref  
+      WHERE description like '%{ECO:%}%'
+  /;
+
+  is_rows_zero($self->dba, $sql_1, $desc_1);
+
+  my $desc_2 = 'No NULL OR Bad Chracters Found In display_label';
+  my $sql_2  = qq/
+      SELECT count(*) FROM xref 
+      WHERE display_label IS NULL
```
james-monkeyshines: This can be removed, it's redundant because the schema disallows NULLs in this field.

lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionDisplayXrefs.pm
```diff
@@ -0,0 +1,77 @@
...
+      OR display_label REGEXP '^[:;\\n\\r\\t~ ]+\$'
```
james-monkeyshines: This regexp doesn't work, for a couple of reasons. First, the double-escaping of n, r and t means that it's looking for those literal characters, rather than whitespace - one escape is all that's needed. Also it's looking for text which consists entirely of these characters, but that doesn't seem sensible - we want text that contains them. But then, I think that would actually be too strict for a critical test - we don't have control over the labels that external sources use, and while some of these seem ill-advised, we may have to tolerate them. The human db has examples of spaces, : and ;
```suggestion
      OR display_label REGEXP '[\n\r\t]+'
```

james-monkeyshines: We could create an advisory datacheck to check for the characters that I suggest removing, but I don't think it's worth it - as these are externally controlled, we can't do much in the event of a failure.

lib/Bio/EnsEMBL/DataCheck/Checks/DescriptionDisplayXrefs.pm
```diff
@@ -0,0 +1,77 @@
...
+  /;
+
+  is_rows_zero($self->dba, $sql_2, $desc_2);
+
+  my $desc_3 = 'No improper import of MIM data.';
```
james-monkeyshines: With the alteration I suggest above, this test becomes redundant and can be removed.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayLabels.pm
```diff
@@ -0,0 +1,67 @@
...
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'DisplayLabels',
+  DESCRIPTION    => 'Check that certain tables have display_labels set',
+  GROUPS         => ['core', 'xref'],
```
james-monkeyshines: This isn't really related to xrefs.
```suggestion
  GROUPS         => ['annotation', 'core'],
```

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayLabels.pm
```diff
@@ -0,0 +1,67 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::DisplayLabels;
```
james-monkeyshines: I think this name is too generic; and most people would assume it's about xrefs, but it's not, it's about prediction transcripts which have a column with the same name as in the xref table. I suggest 'PredictionTranscriptLabels'.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayLabels.pm
```diff
@@ -0,0 +1,67 @@
...
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'DisplayLabels',
+  DESCRIPTION    => 'Check that certain tables have display_labels set',
+  GROUPS         => ['core', 'xref'],
+  DATACHECK_TYPE => 'critical',
```
james-monkeyshines: Might need to test this on current vertebrate dbs and if lots of things fail, might need to be advisory. We're testing for NULL values, which are allowed by the schema - if they really are so bad, why is the field NULLable in the first place?

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayLabels.pm
```diff
@@ -0,0 +1,67 @@
...
+      SELECT count(*) FROM prediction_transcript pt
+      INNER JOIN seq_region sr USING (seq_region_id) 
+      INNER JOIN  coord_system cs USING (coord_system_id)   
+      WHERE cs.species_id = $species_id
+      AND pt.display_label IS NULL
+  /;
+
+  is_rows_zero($self->dba, $sql_1, $desc_1);
+
+  my $desc_2 = 'No NULL values Found in  display_label for simple_feature table';
```
james-monkeyshines: This test is not required, display_label in the simple_feature table is non-NULLable.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXref.pm
```diff
@@ -0,0 +1,101 @@
...
+};
+
+sub tests {
+
+  my ($self) = @_;
+
+  my $species_id = $self->dba->species_id;
+  foreach my $type ("gene", "transcript"){
+
+    my $desc_1 = "${type}s  with valid display_xref_id";
```
james-monkeyshines: This is not critical - most non-vert dbs do not have display xrefs, and new vertebrates might not either; this test needs to be moved to a separate advisory datacheck.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXref.pm
```diff
@@ -0,0 +1,101 @@
...
+    my $sql_1  = qq/
+     SELECT COUNT(*) FROM $type t
+     INNER JOIN seq_region sr USING (seq_region_id) 
+     INNER JOIN  coord_system cs USING (coord_system_id)   
+     WHERE cs.species_id = $species_id
+     AND  t.display_xref_id IS NOT NULL 
+     AND  t.display_xref_id > 0;
```
james-monkeyshines: This '>0' condition is not necessary, other datachecks ensure that this field has a non-zero value.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXref.pm
```diff
@@ -0,0 +1,101 @@
...
+    /;
+
+    is_rows_nonzero($self->dba, $sql_1, $desc_1);
+
+    my $desc_2 = "No ${type}s  with  display_xref_id set to 0";
```
james-monkeyshines: This test can be removed, the ForeignKeys datacheck will fail if any IDs are zero.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXref.pm
```diff
@@ -0,0 +1,101 @@
...
+     SELECT COUNT(*) FROM $type t
+     INNER JOIN seq_region sr USING (seq_region_id) 
+     INNER JOIN  coord_system cs USING (coord_system_id)   
+     WHERE cs.species_id = $species_id     
+     AND t.display_xref_id = 0
+    /;
+
+    is_rows_zero($self->dba, $sql_2, $desc_2);
+
+    my $desc_3 = "No ${type}s have display_xref_ids pointing to non-existent xrefs";
```
james-monkeyshines: Same as previous comment, ForeignKeys datacheck already does this.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXref.pm
```diff
@@ -0,0 +1,101 @@
...
+      AND t.display_xref_id IS NOT NULL 
+      AND xref.xref_id IS NULL;
+    /;
+
+    is_rows_zero($self->dba, $sql_3, $desc_3);
+
+  }
+
+
+    my $desc_4 = "genes have display_xrefs that are from EntrezGene and are numeric";
```
james-monkeyshines: The description doesn't quite capture what we're testing.
```suggestion
    my $desc_4 = "EntrezGene gene names are not numeric";
```

lib/Bio/EnsEMBL/DataCheck/Checks/MartDatasetName.pm
```diff
@@ -60,7 +60,6 @@ sub tests {
   my $dbname = $self->dba->dbc->dbname;
```
james-monkeyshines: This has turned up because these changes were committed to release/101, but the PR is against master.
We want these changes for release 101, so the PR should be against that release, which should then mean this bit goes away...

PR #218, opened 2020-03-23
Healthchecks to datachecks

vinay-ebi:
Added new datachecks:
- GeneDescription
- XrefCigarLines
- XrefVersion


lib/Bio/EnsEMBL/DataCheck/Checks/HGNCTypes.pm
```diff
@@ -0,0 +1,64 @@
...
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'HGNCTypes',
+  DESCRIPTION    => 'Check that HGNC_curated_genes xrefs are on genes, _transcript are on transcript etc',
+  GROUPS         => ['core', 'xref'],
+  DATACHECK_TYPE => 'critical',
+  TABLES         => ['xref'],
+  PER_DB         => 1
```
james-monkeyshines: This isn't a per_db test, because the test query involves the object_xref table; so it needs the extra tables and the species_id filter.

lib/Bio/EnsEMBL/DataCheck/Checks/HGNCTypes.pm
```diff
@@ -0,0 +1,64 @@
...
+};
+
+sub tests {
+  my ($self) = @_;
+  #note these are looking for the *wrong* assignments
+  my %check_type =(
```
james-monkeyshines: I just checked in the production db, and none of these are current (http://production-services.ensembl.org/admin/ensembl_production_db/masterexternaldb/?q=HGNC&is_current=all).
Need to test "HGNC" => "Transcript" and "HGNC_trans_name" => "Gene" (assuming you keep the same logic - I think it'd be clearer if the hash had the expectation of what was correct, and the SQL condition did '<>' rather than '='. That would also test if anything is incorrectly assigned to Translations, which is not currently covered by this test.)

lib/Bio/EnsEMBL/DataCheck/Checks/IdentityXrefCigarLines.pm
```diff
@@ -0,0 +1,53 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::IdentityXrefCigarLines;
```
james-monkeyshines: I think we can probably include this in the XrefFormat datacheck, rather than having its own datacheck.

lib/Bio/EnsEMBL/DataCheck/Checks/XrefVersions.pm
```diff
@@ -0,0 +1,51 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::XrefVersions;
```
james-monkeyshines: This datacheck is not required, it is redundant. There is already a generic datacheck that tests for empty strings in all NULLable fields: https://github.com/Ensembl/ensembl-datacheck/blob/release/101/lib/Bio/EnsEMBL/DataCheck/Checks/BlankNulls.pm

lib/Bio/EnsEMBL/DataCheck/Checks/HGNCNumeric.pm
```diff
@@ -0,0 +1,76 @@
...
+  GROUPS         => ['core', 'xref'],
+  DATACHECK_TYPE => 'critical',
+  TABLES         => ['xref'],
+  PER_DB         => 1
+};
+
+sub tests {
+
+  my ($self) = @_;
+  my $threshold = 0.01;
```
james-monkeyshines: This test can be a lot simpler - the healthcheck was over-complicated, we don't need to bother with any of the threshold stuff, just need a straight test of whether the two fields are the same - if any rows are, then we fail.

lib/Bio/EnsEMBL/DataCheck/Checks/HGNCNumeric.pm
```diff
@@ -0,0 +1,76 @@
...
+       AND x.xref_id=ox.xref_id 
+       AND ox.ensembl_object_type='Gene' 
+       AND ox.ensembl_id=g.gene_id 
+       AND e.db_name LIKE 'HGNC%'
+  /;
+  my $numeric_sql = $all_sql . " AND x.dbprimary_acc=x.display_label";
+  my $all_rows = sql_count($self->dba, $all_sql);
+  if ( $all_rows == 0 ){
+
+      pass( $desc_1);
```
james-monkeyshines: In general, it's best to avoid using the 'pass' and 'fail' methods. It's no longer relevant here, because all this complexity can go, but if you do need a conditional test based on the result of another, you can use the return value of the test, e.g. `my $result = is_rows_zero($self->dba, $all_sql, $desc_1)`

PR #221, opened 2020-04-01
Revert "Also check when the assembly length is too long"

james-monkeyshines:
There is at least one assembly (_Ciona intestinalis_) which violates the stricter condition that was added - this doesn't exactly make sense, it means that there is a "gap" at the end of the sequence. It would be customary to trim those trailing Ns; but the datacheck needs to tolerate existing assemblies where this is not done (since it does not cause any errors, _per se_).

If this stricter condition is useful as a warning, it could be added as an advisory datacheck.


james-monkeyshines:
@MatBarba FYI, I had to revert this change, it's too strict for some vertebrates - please consider an advisory datacheck if the stricter condition is something you want flagged up.

PR #231, opened 2020-04-21
Check for empty strings in set columns

helensch:
General test for set column types. For ENSVAR-3161  to check that variation_feature.consequence_types is not empy string


dglemos:
Tested on a database with empty variation_feature.consequence_types and the test fails - looks ok.

PR #232, opened 2020-04-21
Revert "When running the XrefPrefixes on vertebrates post xrefs, we discovere…"

james-monkeyshines:
The datacheck failures that you saw were valid - there is a genuine problem with these xrefs, for which there is a known bug: https://www.ebi.ac.uk/panda/jira/browse/ENSINT-438.
I have a fix lined up, as per https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-5382.


thomasmaurel:
Arg sorry @james-monkeyshines I completely missed that (Tha'ts why I cced you as a reviewer just in case ;-) ). Thanks for the explanation

PR #238, opened 2020-05-01
Update VersionedGenes.pm

marcoooo:
Added Viruses as versioned expected genes.


lib/Bio/EnsEMBL/DataCheck/Checks/VersionedGenes.pm
```diff
@@ -43,7 +43,7 @@ sub tests {
   my $division = $mca->get_division;
 
   my $version_expected = 0;
-  if ($division eq 'EnsemblVertebrates') {
+  if ($division ~= 'Ensembl(Vertebrates|Viruses)') {
```
james-monkeyshines:
```suggestion
  if ($division =~ /Ensembl(Vertebrates|Viruses)/) {
```

PR #241, opened 2020-05-06
Skip check for EnsemblViruses

helensch:
Skip for division based on approach in PhenotypeMultipleSeqRegions
This is also needed for release/100
Should a separate PR be made on release/100?

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MultipleSeqRegions.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MultipleSeqRegions.pm
@@ -35,6 +35,28 @@ use constant {
   TABLES      => ['structural_variation_feature', 'variation_feature']
 };
 
+sub skip_tests {
+  my ($self) = @_;
+
+  my $desc_dna_dba = 'Core database found';
+  my $dna_dba = $self->get_dna_dba();
+  my $pass = ok(defined $dna_dba, $desc_dna_dba);
+
+  # Note that if $pass is false, we will still execute the 'tests'
+  # method; which is either not required, or will need to done again,
+  # because the datacheck will fail. But it's complicated to do
+  # otherwise, and it doesn't really matter for this datacheck,
+  # because the query runs quickly.
+  if ($pass) {
+    my $mca = $dna_dba->get_adaptor("MetaContainer");
+    my $division = $mca->get_division;
+
+    if ($division eq 'EnsemblViruses') {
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/MultipleSeqRegions.pm
```diff
@@ -35,6 +35,28 @@ use constant {
...
 
+sub skip_tests {
+  my ($self) = @_;
+
+  my $desc_dna_dba = 'Core database found';
+  my $dna_dba = $self->get_dna_dba();
+  my $pass = ok(defined $dna_dba, $desc_dna_dba);
+
+  # Note that if $pass is false, we will still execute the 'tests'
+  # method; which is either not required, or will need to done again,
```
at7, line 46: small typo: will need to **be** done again

at7:
I ran the updated DC on a virus variation database with only one seq_region_id in the variation_feature table and the DC passes as expected with the code update.

lib/Bio/EnsEMBL/DataCheck/Checks/MultipleSeqRegions.pm
helensch: Thanks, fixed

james-monkeyshines:
It makes sense to merge this for now, so as not to hold up the processing of the coronavirus data. But is this datacheck necessary at all? There's already a check that the seq_region table in the variation db matches the one in the core db (https://github.com/Ensembl/ensembl-datacheck/blob/release/101/lib/Bio/EnsEMBL/DataCheck/Checks/ForeignKeysMultiDB.pm#L83). If that passes, does it matter how many seq_regions there are?

helensch:
The check is to flag if there is only a single seq region. This indicates could indicate a failed load.

james-monkeyshines:
My point was that there is nothing inherently wrong with having a single seq_region; and that the ForeignKeysMultiDB duplicates the intended functionality, and is a more comprehensive check - it confirms that the seq_region table in the variation db is consistent with the core db, so would higlight an erroneous single seq_region; and it would also flag up if a load failed after, for example, two (out of 100) seq_regions were added.

helensch:
Yes nothing inherently wrong with having a single seq_region. variation.seq_region is a subset of core db seq_region. If there are multiple seq regions in variation.seq_region then would expect there to be more than distinct seq_region in variation_feature, structural_variation_feature. Please remove the check if it is redundant.

PR #244, opened 2020-05-12
Adding test for the presence of a motif_feature file.

james-monkeyshines:


ilavidas:
Hello James,
If I understand correctly, according to your commit, this check for the Motif Feature file will only be run for species that have a Regulatory Build. At the moment this happens to be true. In the future though, we may have species with a RegBuild but without any motif features or vice versa.
It would probably be better to check whether 'SELECT count(*) FROM motif_feature' is higher than zero. Is it OK if I submit a new PR?

james-monkeyshines:
@ilavidas Yes, please update to fit your use case. Thanks!

PR #252, opened 2020-05-26
Brc4 master

MatBarba:
BRC4/Vectorbase requires a specific set of data checks for its core databases that the "core" group itself doesn't (and shouldn't) cover.

This PR adds a new group "brc4_core" to a subset of the "core" group checks, and adds a couple of new specific checks for BRC4:
* MetaKeyBRC4 for a set of specific meta keys checks
* SeqRegionBRC4 for a set of specific seq_region attributes and synonyms checks
* VersionedGenesBRC4, the opposite of UnversionedGenes (will be replaced by the Ensembl version when it lands)


lib/Bio/EnsEMBL/DataCheck/Checks/VersionedGenesBRC4.pm
```diff
@@ -0,0 +1,83 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
```
marcoooo, line 3: Copyright is wrong.

lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyBRC4.pm
```diff
@@ -0,0 +1,57 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
```
marcoooo, line 3: License should be 2020

MatBarba:
> IMHO, since the need for this group is only for a new metakey to be checked, as far as I am aware. I was not expecting you to "add your group to every single core existing test". Why don't you just run those two group when building your DB ?

Because we also need to exclude some checks (we only want a subset of "core").

lib/Bio/EnsEMBL/DataCheck/Checks/VersionedGenesBRC4.pm
MatBarba: Changed in updated branch

lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyBRC4.pm
MatBarba: Changed in updated branch

marcoooo:
> > IMHO, since the need for this group is only for a new metakey to be checked, as far as I am aware. I was not expecting you to "add your group to every single core existing test". Why don't you just run those two group when building your DB ?
> 
> Because we also need to exclude some checks (we only want a subset of "core").

This is what's struggling me. The core group will be executed for handover, so any test you remove wouldn't pass when you submit your Db. So back to square 1.

That wouldn't solve your issue, right ? Better to remove tests from core group with configuration or meta key value.

MatBarba:
> > > IMHO, since the need for this group is only for a new metakey to be checked, as far as I am aware. I was not expecting you to "add your group to every single core existing test". Why don't you just run those two group when building your DB ?
> > 
> > 
> > Because we also need to exclude some checks (we only want a subset of "core").
> 
> This is what's struggling me. The core group will be executed for handover, so any test you remove wouldn't pass when you submit your Db. So back to square 1.
> 
> That wouldn't solve your issue, right ? Better to remove tests from core group with configuration or meta key value.

Those dbs are not to be handed over to Ensembl, they are expected to be transferred to our partners in UPenn, so we have different needs.
We've already discussed this with James (cf Slack prod) and he indicated that the best option is to create a group, instead of division-specific configurations. Is there a new opinion on this?

lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyBRC4.pm
```diff
@@ -0,0 +1,57 @@
...
+  DB_TYPES       => ['core'],
+  TABLES         => ['meta']
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my @optional = qw/
+  assembly.accession
+  species.taxonomy_id
```
james-monkeyshines, line 42: This one is mandatory in all core dbs.

lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyBRC4.pm
```diff
@@ -0,0 +1,57 @@
...
+  DESCRIPTION    => 'Expected meta keys for BRC4 cores',
+  GROUPS         => ['brc4_core'],
+  DB_TYPES       => ['core'],
+  TABLES         => ['meta']
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my @optional = qw/
```
james-monkeyshines, line 40:
```suggestion
  my @expected = qw/
```
Parameter name has been copied from another datacheck, could be misleading in this case.

lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyBRC4.pm
```diff
@@ -0,0 +1,57 @@
...
+  assembly.accession
+  species.taxonomy_id
+  BRC4.component
+  BRC4.organism_abbrev
+  /;
+
+  my $mca = $self->dba->get_adaptor("MetaContainer");
+
+  foreach my $meta_key (@optional) {
```
james-monkeyshines, line 49:
```suggestion
  foreach my $meta_key (@expected) {
```
To match the earlier suggestion.

lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionBRC4.pm
```diff
@@ -0,0 +1,171 @@
...
+sub check_seq_attrib_name {
+  my ($self, $species_id, $attrib_code) = @_;
+
+  my $desc = "All seq_regions have a '$attrib_code' attribute";
+  my $diag = "No attrib_name '$attrib_code' for seq_region";
+  my $sql  = qq/
+    SELECT sr.name
+    FROM seq_region sr
+      INNER JOIN coord_system cs USING (coord_system_id)
+      INNER JOIN
```
james-monkeyshines, line 83: Can you not remove the nested query here, since we have a single coord_system?

lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionBRC4.pm
```diff
@@ -0,0 +1,171 @@
...
+sub check_seq_attrib_name_coord {
+  my ($self, $species_id, $coord_id, $attrib_code) = @_;
+
+  my $desc = "All seq_regions have a '$attrib_code' attribute";
+  my $diag = "No attrib_name '$attrib_code' for seq_region";
+  my $sql  = qq/
+    SELECT sr.name
+    FROM seq_region sr
+      INNER JOIN coord_system cs USING (coord_system_id)
+      INNER JOIN
```
james-monkeyshines, line 115: As with previous comment, I think this subquery is unnecessary - and wrong in this case, since one of the attributes you are testing, on line 71, is whether 'toplevel' is annotated...

lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionBRC4.pm
```diff
@@ -0,0 +1,171 @@
...
+
+sub tests {
+  my ($self) = @_;
+
+  my $species_id = $self->dba->species_id;
+
+  $self->check_seq_attrib_name($species_id, 'BRC4_seq_region_name');
+
+  # Check for INSDC accession (not systematic in case out assembly is not in sync with INSDC)
+  #$self->check_seq_synonym($species_id, 'INSDC');
```
james-monkeyshines, line 46: This could be moved to a separate 'advisory' datacheck. (But in MetaKeyBRC4 you have 'assemby.accession' as a required meta_key, which implies that the sequences must be in sync?)

james-monkeyshines:
Regarding the need to add the new group to all of the datachecks - currently that's the best thing to do. The idea of the 'core' group is that it checks everything that is essential for a core database; so in theory, Marc's idea is correct, for a given database we should apply the 'core' group, plus any groups that contain datachecks for a specific team. But, actually, 'core' also enforces some Ensembl conventions, which are not always applicable, so a complete alternative core group is required (we could run into similar issues with Gramene and ParaSite).
An alternative approach would be to reformulate the 'core' group, and split off any Ensembl-only ones into a new 'core_ensembl' group. This might turn out to be cleaner, but we'd need to weigh up the pros and cons.

marcoooo:
> > > > IMHO, since the need for this group is only for a new metakey to be checked, as far as I am aware. I was not expecting you to "add your group to every single core existing test". Why don't you just run those two group when building your DB ?
> > > 
> > > 
> > > Because we also need to exclude some checks (we only want a subset of "core").
> > 
> > 
> > This is what's struggling me. The core group will be executed for handover, so any test you remove wouldn't pass when you submit your Db. So back to square 1.
> > That wouldn't solve your issue, right ? Better to remove tests from core group with configuration or meta key value.
> 
> Those dbs are not to be handed over to Ensembl, they are expected to be transferred to our partners in UPenn, so we have different needs.
> We've already discussed this with James (cf Slack prod) and he indicated that the best option is to create a group, instead of division-specific configurations. Is there a new opinion on this?

No no, I keep with what we (James) said on slack. I was thinking your db was intended to be handover to us. The meta key question was more to enable/disable check on geneset (for the use case when only assembly are in DB, I believe in this case, DB would eventualy be handed over to us)

marcoooo:
> Regarding the need to add the new group to all of the datachecks - currently that's the best thing to do. The idea of the 'core' group is that it checks everything that is essential for a core database; so in theory, Marc's idea is correct, for a given database we should apply the 'core' group, plus any groups that contain datachecks for a specific team. But, actually, 'core' also enforces some Ensembl conventions, which are not always applicable, so a complete alternative core group is required (we could run into similar issues with Gramene and ParaSite).
> An alternative approach would be to reformulate the 'core' group, and split off any Ensembl-only ones into a new 'core_ensembl' group. This might turn out to be cleaner, but we'd need to weigh up the pros and cons.

Completely agreed. We might end up with some `core` group dedicagted to `team` our `site`. But this need to be discussed first. So ok for adding new group.

lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionBRC4.pm
MatBarba: That's a problem for us when the assembly has an INSDC accession, but the sequences are actually different to what is currently in INSDC. This can happen when we have made updates to the assembly but haven't (yet) pushed the changes back to INSDC. I guess this should be an advisory check still, because this is not something we want to keep as is.

lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionBRC4.pm
MatBarba: Done with last commit

lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionBRC4.pm
MatBarba: True, good catch. Fixed in the last commit.

PR #254, opened 2020-06-04
Detect CAFE using an mlss_tag

carlacummins:
As it stands, CAFE is decided on a per-division basis only. 

This PR updates the DC to also check the method_link_species_set_tag table to decide whether gain/loss data exists on a per-mlss basis.


carlacummins:
Thank you - I will use the GOC trick. I had removed `skip_tests` and then realised in my test run that the test failed when there were no mlsses with the `has_cafe` tag, so I added it back. 😅

PR #255, opened 2020-06-10
Added new module DataCheckTapToJson to convert datacheck tap output t…

vinay-ebi:
Converting Datacheck output TAP format to json,  helps to render as html in datacheck service.
The output is parsed by species.


james-monkeyshines:
@thomasmaurel Since I've made quite a lot of changes to the original PR I'm not the best person to approve, please can you give it a once-over...

PR #258, opened 2020-06-10
"ancestral_sequences" is what the Compara API uses to link GenomeDBs to Core DBAdaptors

muffato:
Not really a bug, but in case we want to check the ancestral database against a compara database, `ancestral_sequences` is the best way of linking both (we have such a check as a separate script that we might consider porting to the DC framework).
`Ancestral sequences` is indeed the default Registry name for Vertebrates, but it only works because `ancestral_sequences` is added as an alias.


james-monkeyshines:
Thanks @muffato!

PR #265, opened 2020-06-22
remove known phenotype descriptions

ima23:
Updating known and accepted phenotype descriptions in check.

Before this change the DC would fail if 'none' would be present as a phenotype description. This is accepted as long as the phenotype class attrib is 'non_specified'.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/PhenotypeDescription.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/PhenotypeDescription.pm
@@ -75,7 +75,7 @@ sub tests {
   my $sql_non_term = qq/
       SELECT phenotype_id
       FROM phenotype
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/PhenotypeDescription.pm
```diff
@@ -75,7 +75,7 @@ sub tests {
   my $sql_non_term = qq/
       SELECT phenotype_id
       FROM phenotype
-      WHERE lower(description) in ("none", "not provided", "not specified", "not in omim", "variant of unknown significance", "not_provided", "clinvar: phenotype not specified", "?", ".")
+      WHERE lower(description) in ( "?", ".")
```
helensch, line 78: Is lower still needed?

ima23: Yes it is.

PR #266, opened 2020-06-29
Sort transcripts to use the dna cache

MatBarba:
ProteinTranslation can take a very long time on long sequences.
Sorting the transcripts by seq_region and start ensures the dna sequence
cache is used more effectively.

Example of speedup: from 1.5h to 30min on Aedes aegypti


james-monkeyshines:
Excellent, thanks @MatBarba!

PR #269, opened 2020-07-01
Move test out of ForeignKeysCompara to new DC

CristiGuijarro:
## CheckSpeciesTrees

We only detected in `e101` that the species tree in vertebrates had not been merged when the `ForeignKeysCompara` datacheck was run prior to handover. Here we have a species tree specific DC that checks that all expected species trees with linked mlsss are present in the release database. The FK check for the species tree has been moved here too, so that it is not duplicated.

There is some overlap in the tests within `CheckSpeciesTree`, happy to discuss. Was aiming to provide more specific descriptions in the non-fk derived tests to make failures easier to understand.

### Jira ticket linked:
- [Add a datacheck to check that all species-tree are present in the database](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-3602)

## ForeignKeysCompara

`ForeignKeysCompara` was not quite compatible with embedding in our pipelines because it expects all species_sets must be used by an mlss, this is not the case in our production pipeline databases. A new method to detect if a database is an ehive pipeline database or not was placed in `Utils` provisionally, if there is somewhere better to do this or an equal and already implemented method, open to those suggestions. Will add test for coverage if agreed on.

### Jira ticket linked:
- [Update the ForeignKey datacheck to make it pass on production databases](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-3606)

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSpeciesTrees.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSpeciesTrees.pm
@@ -83,7 +83,7 @@ sub tests {
         (class LIKE 'GenomicAlignTree%' OR class LIKE '%multiple_alignment')
     )
   /;
-  fk($self->dba, 'species_tree_root',       'method_link_species_set_id', 'method_link_species_set', 'method_link_species_set_id', $species_tree_root_constraint);
+  fk($dba, 'species_tree_root', 'method_link_species_set_id', 'method_link_species_set', 'method_link_species_set_id', $species_tree_root_constraint);
 }
 
 1;
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSpeciesTrees.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSpeciesTrees.pm
@@ -73,7 +73,7 @@ sub tests {
   # Check that all the species_trees correspond to a method_link_species_set that should have a species tree
   my $desc_2 = "All the species_trees with corresponding method_link_species_sets are expected";
   my $row_count_sql = "SELECT COUNT(*) FROM species_tree_root";
-  cmp_rows($dba, $row_count_sql, "==", $expected_tree_count, $desc_2);
+  is_rows($dba, $row_count_sql, $expected_tree_count, $desc_2);
 
   my $species_tree_root_constraint = q/
     method_link_id IN (
@@ -87,4 +87,3 @@ sub tests {
 }
 
 1;
-
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSpeciesTrees.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckSpeciesTrees.pm
@@ -1,0 +1,90 @@
+=head1 LICENSE
+
+Copyright [2018-2020] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckSpeciesTrees;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use Bio::EnsEMBL::Utils::SqlHelper;
+use Data::Dumper;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'CheckSpeciesTrees',
+  DESCRIPTION    => 'The expected number of species trees have been merged',
+  GROUPS         => ['compara', 'compara_gene_trees', 'compara_genome_alignments'],
+  DATACHECK_TYPE => 'critical',
+  DB_TYPES       => ['compara'],
+  TABLES         => ['method_link_species_set', 'species_set', 'species_tree_root']
+};
+
+sub tests {
+  my ($self) = @_;
+  my $dba    = $self->dba;
+  my $helper = $dba->dbc->sql_helper;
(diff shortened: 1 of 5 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CheckSpeciesTrees.pm
```diff
@@ -0,0 +1,90 @@
...
+
+  my $species_tree_root_constraint = q/
+    method_link_id IN (
+      SELECT method_link_id FROM method_link
+      WHERE
+        method_link_id < 100 AND
+        (class LIKE 'GenomicAlignTree%' OR class LIKE '%multiple_alignment')
+    )
+  /;
+  fk($dba, 'species_tree_root', 'method_link_species_set_id', 'method_link_species_set', 'method_link_species_set_id', $species_tree_root_constraint);
```
muffato, line 86: I know this is an exact copy of the original `fk` check, but looking into it, I wonder if that's right.

My understanding of how `fk` works is that the additional constraint should apply to the first table, not the second one. See https://github.com/Ensembl/ensembl-datacheck/blob/release/100/lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm#L336-L378 : `fk` does `WHERE t1.$col1 IS NOT NULL AND t2.$col2 IS NULL`, so all the rows in `t2` have to be NULLs, so none can't match the condition. In summary, this `fk` (and other similar ones in ForeignKeysCompara) do nothing ? Can anyone double-check this ?

CristiGuijarro: I think you are right, that it is not right. I am trying to get my head around the fact that the `LEFT JOIN` is on `method_link_species_set_id` in both tables, the `IS NULL` / `IS NOT NULL` would not work because the `method_link_id` is from the second table, which has to be `IS NULL`- which means it is not there? Is that similar to your thinking? The check in this case looks like:
`    SELECT * FROM
    species_tree_root t1 LEFT JOIN method_link_species_set t2 ON t1.method_link_species_set_id = t2.method_link_species_set_id;
    WHERE t1.method_link_species_set_id IS NULL AND t2.method_link_species_set_id IS NOT NULL AND 
    method_link_id IN (
      SELECT method_link_id FROM method_link
      WHERE
        method_link_id < 100 AND
        (class LIKE 'GenomicAlignTree%' OR class LIKE '%multiple_alignment')
    );`

muffato: Yep that's my thinking.

Ah right, good idea. It's actually easier with the actual query:
```
ensro@mysql-ens-compara-prod-1.ebi.ac.uk:4485/ensembl_compara_101 > EXPLAIN SELECT * FROM species_tree_root t1 LEFT JOIN method_link_species_set t2 ON t1.method_link_species_set_id = t2.method_link_species_set_id WHERE t1.method_link_species_set_id IS NULL AND t2.method_link_species_set_id IS NOT NULL AND method_link_id IN ( SELECT method_link_id FROM method_link WHERE method_link_id < 100 AND (class LIKE 'GenomicAlignTree%' OR class LIKE '%multiple_alignment') );
+----+-------------+-------+------+---------------+------+---------+------+------+------------------+
| id | select_type | table | type | possible_keys | key  | key_len | ref  | rows | Extra            |
+----+-------------+-------+------+---------------+------+---------+------+------+------------------+
|  1 | SIMPLE      | NULL  | NULL | NULL          | NULL | NULL    | NULL | NULL | Impossible WHERE |
+----+-------------+-------+------+---------------+------+---------+------+------+------------------+
1 row in set (0.033 sec)
```

MySQL is pretty clear about what it thinks of the `WHERE` :)

muffato: If we come back to the original HCs:
- https://github.com/Ensembl/ensj-healthcheck/blob/release/100/src/org/ensembl/healthcheck/testcase/compara/ForeignKeyMLSSIdGenomic.java#L45-L59 and https://github.com/Ensembl/ensj-healthcheck/blob/release/100/src/org/ensembl/healthcheck/testcase/compara/ForeignKeyMLSSIdHomology.java#L59-L60
- https://github.com/Ensembl/ensj-healthcheck/blob/release/100/src/org/ensembl/healthcheck/testcase/compara/AbstractMLSSIdToData.java#L30-L41

Two sorts of checks were performed on the `species_tree_root` table:
1. All the MLSSs that match the constraint have some data in the `species_tree_root` table
2. `species_tree_root.method_link_species_set_id` are valid mlss_ids

Check n°2 is actually a real SQL foreign key check that is encoded in our schema and captured by ForeignKeysCompara: https://github.com/Ensembl/ensembl-compara/blob/release/100/sql/table.sql#L469

Check n°1 is actually what you have implemented, but with method_link types instead of classes, which is fine.

So I think you could delete this block and it'd be OK ?
```suggestion
```

And we need to create a ticket to fix all the `fk` calls of ForeignKeysCompara that have a condition on the wrong table.

lib/Bio/EnsEMBL/DataCheck/Utils.pm
```diff
@@ -212,4 +212,29 @@ sub hash_diff {
   return (\%diff);
 }
 
+=item B<is_ehive_db>
+
+is_ehive_db($dba);
```
james-monkeyshines, line 217: I'm a little cautious about this change, because it could mean that failures go undetected - for example, if you run the datacheck and it passes (with some code skipped because it's in pipeline mode), the history file will record a successful run. When you subsequently remove the hive tables, and if you don't change the database name, the datacheck will skip due to the history, so the tests within the conditional would not get run. In the particular use case in this PR, I don't think this would be a problem, due to the db name change in the merging process.
But in principle, introducing different behaviour for embdded datachecks could be storing up trouble for later...
That said, I understand the reason for you introducing this function, and only the compara team create databases in this manner; so if you think it is a manageable risk, it's fine to add this method, but I would ask that you change the method name to `is_compara_ehive_db`, so that other teams know this is not something they need to concern themselves with, and do not get confused about datachecks running differently when embedded in pipelines. (An alternative approach with less potential for future harm, if you envision this sort of situation as rare, might be to have this function in the ForeignKeysCompara datacheck, rather than Utils. Or, if you do prefer not to use this method, the 'reverse-direction' FK tests could be put in a separate datacheck, which isn't included in the pipeline set.)

CristiGuijarro: I think I will change the name in the way you have suggested for now, I have another ticket regarding `ForeignKeysCompara` in which I will be able to consider the use further in another PR. Would that be okay?

james-monkeyshines: Sure, thanks Cristi.

PR #274, opened 2020-07-14
Fix PhenotypeMultipleSeqRegions division fetch

ima23:


ima23:
Tagging @helensch as agreed. Thanks.

ima23:
Thanks @james-monkeyshines for checking and thanks @helensch for digging into the edge case and finding a work around.
I've reverted the PR code to what was before and only added a check if the core mca is undef for clearer error messaging.

lib/Bio/EnsEMBL/DataCheck/Checks/PhenotypeMultipleSeqRegions.pm
```diff
@@ -59,8 +59,11 @@ sub skip_tests {
   # because the query runs quickly.
   if ($pass) {
     my $mca = $dna_dba->get_adaptor("MetaContainer");
-    my $division = $mca->get_division;
+    if (! defined $mca) {
+      die("no mca")  # Make a better die message!
```
helensch, line 63: Can the die message be updated and the comment removed.

ima23: Yes, thanks. I updated it below.

ima23:
Thanks @james-monkeyshines. Indeed there looks like something is fishy there. The issue was highlighted when using a core species name in the registry starting with uppercase. Replicated now on human staging 102 dbs. It runs when the species is lowercase. The rest of DCs seem fine with upper or lower case species name.
Yes, the fail message is that it can't call  method 'get_division'  on an undefined value as $mca is not defined for some reason.
The work around is to make sure species name starts with lowercase before running the DCs.

james-monkeyshines:
OK, I think that this is because a name with a capital letter would be an alias, rather than the production_name - using that would work in most cases due to registry magic, but when you want to connect to a related db, various registry manipulations are needed. And those don't work with aliases. This bit of the datacheck code is already pretty fiendish, so I don't want to complicate it further by dealing with aliases - I'll add a note to the docs that any passed 'species' parameters should be the production_name.

PR #281, opened 2020-07-24
Check for ProteinFunctionPredictions table

helensch:
Check for NULL values as found in earlier releases.
Will review schema for protein_function_predictions

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/index.json
+++ b/lib/Bio/EnsEMBL/DataCheck/index.json
@@ -1806,7 +1806,7 @@
    },
    "ProteinFunctionPredictions" : {
       "datacheck_type" : "critical",
-      "description" : "prediction_matrix is not NULL",
+      "description" : "prediction_matrix is not NULL or empty",
       "groups" : [
          "variation"
       ],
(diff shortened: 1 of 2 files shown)


helensch:
@dglemos please could you have a look at

lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -1804,6 +1804,15 @@
       "name" : "ProteinFeatures",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::ProteinFeatures"
    },
+   "ProteinFunctionPredictions" : {
+      "datacheck_type" : "critical",
+      "description" : "prediction_matrix is not NULL",
```
dglemos, line 1809: Description should be 'prediction_matrix is not NULL or empty'.

helensch:
Thanks @james-monkeyshines and @dglemos. index.json has been updated. As noted will review the  schema for protein_function_predictions as agree that it would be better in schema and be picked up by BlankNulls. When the schema is changed, will remove this redudant check.

PR #284, opened 2020-08-06
Reinstate motif_feature_variation check

james-monkeyshines:
The naive comparison was intractable, due to the very large number of motif_features for human. This update uses batches, attempting to retrieve the motif_feature_variation stable IDs from the fungen db (via the API). For human, with ~5 million IDs, this takes ~5 minutes.

Regulation team also advised that motif_features are not dependent on having a regulatory build, so removed the skip block (could have moved it further down the module, but the benefit of that would be minimal, since we've already connected to the funcgen db for the motif_feature test, so compete removal simplifies the code.)

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/ForeignKeysMultiDB.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/ForeignKeysMultiDB.pm
@@ -39,7 +39,7 @@ sub tests {
   my ($self) = @_;
 
   if ($self->dba->group eq 'variation') {
-    $self->variation_core_fk();
+    #$self->variation_core_fk();
     $self->variation_funcgen_fk();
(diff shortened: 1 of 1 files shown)


james-monkeyshines:
> @james-monkeyshines what is the recommended memory required for this test for human?
> Running on mouse it took 20 minutes with 8GB.

I didn't remember needing that much memory - I just ran it for human and then mouse, and it used just under 1GB for human, and 1.5GB for mouse.

helensch:
Thanks @james-monkeyshines. Sorry it did not use 8GB. Mouse failed with default of 1GB and I submitted with higher memory. I had similar memory usage as you (under 1GB for human and 1.5GB for mouse). My runs are taking a bit longer than yours (mouse 2 runs: 19min, 16min, human 2 runs: 22min, 6min). Both times are good. Thanks for re-instating test so we no longer need to run health check.

PR #285, opened 2020-08-06
Cross-database compara datachecks

james-monkeyshines:
The DNAFragCore datacheck replaces the compara.CheckTopLevelDnaFrag and the eg_compara.ControlledTableDnafrag healthchecks. It checks for consistency between the dnafrag and seq_region tables. If the number of seq_regions is < 10,000, then names, lengths, and reference status are checked - otherwise only the counts are compared. This is for tractability - currently the datacheck takes ~20 mins for vertebrates (300-odd species). Without this condition, it takes ~6 hours. Obviously this constraint could be removed, if necessary; and it could potentially be speeded up with more complicated code, but when a species has half a million toplevel sequences, there's only so much you can do...

The GenomeDBCore datacheck replaces the compara.CheckGenomeDB healthcheck. It checks for consistency between the genome_db and core meta tables.

The ControlledTablesCompara datacheck (mostly) replaces the eg_compara.ControlledTable* healthchecks. It includes tests that compare the compara tables to those in the compara_master; so requires a compara_master database in the registry, with the 'species' parameter the same as the database name. It does not require the tables to be identical, rather the compara should be a subset of the compara_master. It also checks the compara ncbi_taxa_node table against the ncbi_taxonomy database (again, the latter needs to be in the registry) - this is a very large table, so the comparison is done with a checksum, because the expectation is an exact match. (The ncbi_taxa_name table is not checked, because the compara db table has extra rows - but just checking ncbi_taxa_node should be enough to test whether we're up-to-date).

While adding these new datachecks, I found a few minor bugs in ControlledTablesCore, ControlledTablesVariation, and Utils.pm, so I've included them here, to save on a further PR...

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCompara.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCompara.pm
@@ -91,13 +91,7 @@ sub master_tables {
 sub consistent_data {
   my ($self, $helper, $master_helper, $table) = @_;
 
-  # We need things returned in a consistent order, for which we need
-  # columns names. Easiest way to get them is to return one row.
-  my $row_sql  = "SELECT * FROM $table LIMIT 1";
-  my @row = @{ $helper->execute(-SQL => $row_sql, -use_hashrefs => 1) };
-  my $columns = join(", ", keys %{$row[0]});
-
-  my $sql = "SELECT * FROM $table ORDER BY $columns";
+  my $sql = "SELECT * FROM $table $sql_filter";
   my @data =
     @{ $helper->execute(-SQL => $sql, -use_hashrefs => 1) };
   my @master_data =
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCompara.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCompara.pm
@@ -33,7 +33,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME           => 'ControlledTablesCompara',
   DESCRIPTION    => 'Controlled tables are consistent with compara master database',
-  GROUPS         => ['controlled_tables', 'compara', 'compara_master'],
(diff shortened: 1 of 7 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCompara.pm
```diff
@@ -0,0 +1,174 @@
...
+  my ($self, $helper, $tables) = @_;
+
+  # This test requires a compara_master database in the registry,
+  # where the 'species' parameter is the same as the database name.
+  my $desc_1 = "Compara master database found";
+  my $master_db_name = 'ensembl_compara_master';
+  my $division = $self->dba->get_division;
+  if ($division ne 'vertebrates') {
+    $master_db_name =~ s/ensembl/$division/;
+  }
```
muffato, line 73: In Compara we have a different naming scheme. We name our master databases `ensembl_compara_master(_$DIVISION)`, but they are always named `compara_master` in our Registry config files to simplify cross-divisions work.

I could make this DC try both schemes ? Otherwise, we could add a command-line argument in `run_datachecks.pl` that holds the URI of the master database (like `old_server_uri`).

muffato:
A complete run of DCs for Vertebrates already take >1.8 days. 6 hours is a lot, yes, but that's the price to pay to actually be safe because the tables are so big. I'm not comfortable skipping a test because it takes too long.

lib/Bio/EnsEMBL/DataCheck/Checks/ControlledTablesCompara.pm
james-monkeyshines: If the convention is to name it `compara_master` we can just go with that, and try to keep things simple. As long as the non-vert teams are aware of it, they can provide an appropriate registry.

james-monkeyshines:
> A complete run of DCs for Vertebrates already take >1.8 days. 6 hours is a lot, yes, but that's the price to pay to actually be safe because the tables are so big. I'm not comfortable skipping a test because it takes too long.

OK, fair enough.

james-monkeyshines:
Thanks for your input @muffato. If this is now all good from your side, please can you approve https://github.com/Ensembl/ensembl-datacheck/pull/294 and I'll get it merged (because we have protected branches, the easiest way to resolve the merge conflicts was on a fresh PR).

PR #287, opened 2020-08-07
Updating xref_mapping group

james-monkeyshines:


marcoooo:
I guess this is related to the discussion you had with Mag about Xref tests? If so can you point from this PR to the one related in the production repo, if  any?

james-monkeyshines:
Xref pipeline config updated: https://github.com/Ensembl/ensembl-production/pull/460

PR #294, opened 2020-08-25
Cross-database compara datachecks + conflict resolution

james-monkeyshines:
Copy of #285 that resolves merge conflicts.


muffato:
OK. I'll push my changes related to DNAFragCore here

muffato (approved):
I have pushed all my changes here and in ensembl-compara. I have also piggybacked a fix in CheckGenomicAlignments (cc @CristiGuijarro ).

The updated DNAFragCore now runs on all species, regardless of the number of regions, but it's much quicker as I have made sure the data are fetched in a small number of queries. The entire set of Vertebrates can be checked in 30 min

james-monkeyshines (commented):
Nifty!

PR #295, opened 2020-08-25
Update Ontology Tests

radome:
- CompareOntologyTerm to test each ontology namespace alone
- Add CheckOntologyTerm to test presence of critical accession IDs

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckOntologyTerm.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckOntologyTerm.pm
@@ -1,0 +1,54 @@
+=head1 LICENSE
+
+Copyright [2018-2020] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::CheckOntologyTerm;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'CheckOntologyTerm',
+  DESCRIPTION    => 'Check presence of critical terms',
+  GROUPS         => ['ontologies'],
+  DATACHECK_TYPE => 'critical',
+  DB_TYPES       => ['ontology'],
+  TABLES         => ['term'],
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $desc = 'Critical terms are present in '.$self->dba->dbc->dbname;
+  my @critical_terms = (
+      "'EFO:0003900'"
+  );
+  my $len = @critical_terms;
+
+  my $sql  = "SELECT COUNT(*) FROM term WHERE accession IN (".join(", ", @critical_terms).")";
+
+  is_rows($self->dba, $sql, $len, $desc);
+}
+
+1;
+
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckOntologyTerm.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckOntologyTerm.pm
@@ -45,7 +45,7 @@ sub tests {
   );
   my $len = @critical_terms;
 
-  my $sql  = "SELECT COUNT(*) FROM term WHERE accession IN (".join(", ", @critical_terms).")";
+  my $sql = "SELECT COUNT(*) FROM term WHERE accession IN (".join(', ', @critical_terms).")";
 
   is_rows($self->dba, $sql, $len, $desc);
 }
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckOntologyTerm.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckOntologyTerm.pm
@@ -39,7 +39,7 @@ use constant {
 sub tests {
   my ($self) = @_;
 
-  my $desc = 'Critical terms are present in '.$self->dba->dbc->dbname;
+  my $desc = 'Critical terms are present in term table';
   my @critical_terms = (
       "'EFO:0003900'"
   );
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CompareOntologyTerm.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CompareOntologyTerm.pm
@@ -31,7 +31,7 @@ use constant {
   NAME           => 'CompareOntologyTerm',
   DESCRIPTION    => 'Compare Term counts between current and previous ontology database',
   GROUPS         => ['ontologies'],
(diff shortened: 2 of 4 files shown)


lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -278,16 +278,33 @@ sub row_subtotals {
...
+    my $len_elem2 = @{@$rows2[0]};
+    if ($len_elem2 != 2) {
+      die "Invalid SQL query for row_subtotals. Must return exactly two columns, a key and a number.\n($sql2)"
+    }
+    else {
+      my $elem2 = @$rows2[0];
+      die "Invalid SQL query for row_subtotals. Second column must be a number.\n($sql2)" unless @$elem2[1] =~ m/[0-9]+/;
+    }
+  }
+
```
radome, line 307: @james-monkeyshines There's probably is a better way to check length and type of the query returned elements, but my knowledge of Perl is quite limited so I'm very open to suggestions 🙂

lib/Bio/EnsEMBL/DataCheck/Checks/CompareOntologyTerm.pm
```diff
@@ -44,10 +44,15 @@ sub tests {
...
              $self->dba->dbc->dbname.' compared to '.$old_dba->dbc->dbname;
-  my $sql  = 'SELECT COUNT(*) FROM term';
-  row_totals($self->dba, $old_dba, $sql, undef, 1.00, $desc);
+  my $sql  = q/
+    SELECT CONCAT(ontology.name, ':', ontology.namespace), COUNT(*)
+    FROM term
+    INNER JOIN ontology ON term.ontology_id=ontology.ontology_id
+    GROUP BY ontology.ontology_id
+  /;
+  row_subtotals($self->dba, $old_dba, $sql, undef, 1.00, $desc);
```
marcoooo, line 55: @james-monkeyshines is the 1.00 meaning "I will fail if the difference is 100%"

radome: From `row_subtotals()` docs:
```
row_subtotals($dbc1, $dbc2, $sql1, $sql2, $min_proportion, $test_name)

By default, the test only passes if the counts are exactly the same. To allow
for some wiggle room, C<$min_proportion> can be used to define the minimum
acceptable difference between the counts. For example, a value of 0.75 means
that the count for C<$dbc1> must not be less than 75% of the count for C<$dbc2>.
If two SQL statements are given, a value of 0.75 means that the count for
C<$sql1> must not be less than 75% of the count for C<$sql2>.
```

marcoooo: :+1:

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
james-monkeyshines: This approach is fine, treating the array as a scalar to get the count is the Perlish way to do it (and this is much nicer than the former parsing of the SQL query). But the dereferencing of the array elements is maybe a little non-standard (although it works fine), so I'll make a couple of suggestions there...

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -278,16 +278,33 @@ sub row_subtotals {
...
-  }
-
   my ( undef, $rows1 ) = _query( $dbc1, $sql1 );
   my ( undef, $rows2 ) = _query( $dbc2, $sql2 );
 
+  my $len1 = @$rows1;
+  my $len2 = @$rows2;
+
+  if ($len1 > 0) {
+    my $len_elem1 = @{@$rows1[0]};
```
james-monkeyshines, line 288:
```suggestion
    my $len_elem1 = @{$$rows1[0]};
```

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -278,16 +278,33 @@ sub row_subtotals {
...
+    if ($len_elem1 != 2) {
+      die "Invalid SQL query for row_subtotals. Must return exactly two columns, a key and a number.\n($sql1)"
+    }
+    else {
+      my $elem1 = @$rows1[0];
+      die "Invalid SQL query for row_subtotal. Second column must be a number.\n($sql1)" unless @$elem1[1] =~ m/[0-9]+/;
```
james-monkeyshines, line 294:
```suggestion
      my $count1 = $$rows1[0][1];
      die "Invalid SQL query for row_subtotal. Second column must be a number.\n($sql1)" unless $count1 =~ /^[0-9]+$/;
```

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -278,16 +278,33 @@ sub row_subtotals {
...
+    }
+  }
+  if ($len2 > 0) {
+    my $len_elem2 = @{@$rows2[0]};
```
james-monkeyshines, line 298:
```suggestion
    my $len_elem2 = @{$$rows2[0]};
```

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -278,16 +278,33 @@ sub row_subtotals {
...
+    if ($len_elem2 != 2) {
+      die "Invalid SQL query for row_subtotals. Must return exactly two columns, a key and a number.\n($sql2)"
+    }
+    else {
+      my $elem2 = @$rows2[0];
+      die "Invalid SQL query for row_subtotals. Second column must be a number.\n($sql2)" unless @$elem2[1] =~ m/[0-9]+/;
```
james-monkeyshines, line 304:
```suggestion
      my $count2 = $$rows2[0][1];
      die "Invalid SQL query for row_subtotals. Second column must be a number.\n($sql2)" unless $count2 =~ /^[0-9]+$/;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckOntologyTerm.pm
```diff
@@ -0,0 +1,54 @@
...
+  GROUPS         => ['ontologies'],
+  DATACHECK_TYPE => 'critical',
+  DB_TYPES       => ['ontology'],
+  TABLES         => ['term'],
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $desc = 'Critical terms are present in '.$self->dba->dbc->dbname;
```
james-monkeyshines, line 42: It doesn't matter a great deal, but note that the database name is included automatically in the output, so you don't need it in the description, e.g. it'll be something like:
```
# Subtest: CheckOntologyTerm
    # Subtest: multi, ontology, ensembl_ontology
        ok 1 - Critical terms are present in ensembl_ontology
```

lib/Bio/EnsEMBL/DataCheck/Checks/CompareOntologyTerm.pm
```diff
@@ -31,7 +31,7 @@ use constant {
   NAME           => 'CompareOntologyTerm',
   DESCRIPTION    => 'Compare Term counts between current and previous ontology database',
   GROUPS         => ['ontologies'],
-  DATACHECK_TYPE => 'advisory',
+  DATACHECK_TYPE => 'critical',
   DB_TYPES       => ['ontology'],
   TABLES         => ['term']
```
james-monkeyshines, line 36:
```suggestion
  TABLES         => ['ontology', 'term']
```
Since we're now joining to the ontology table, want changes there to trigger re-runs of the datacheck.

james-monkeyshines (changes requested):

radome:
Good points @james-monkeyshines, thank you!

radome:
Closing this as I should've PR against `master`.

PR #298, opened 2020-09-02
Update Ontology Tests & row_subtotals

radome:
- CompareOntologyTerm to test each ontology namespace alone
- Add CheckOntologyTerm to test presence of critical accession IDs

Cherry Pick of #295

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
@@ -281,6 +281,14 @@ sub row_subtotals {
   my ( undef, $rows1 ) = _query( $dbc1, $sql1 );
   my ( undef, $rows2 ) = _query( $dbc2, $sql2 );
 
+  if (not defined $rows1) {
+    die "Invalid SQL query for row_subtotals.\n($sql1)";
+  };
+
+  if (not defined $rows2) {
+    die "Invalid SQL query for row_subtotals.\n($sql2)";
+  };
+
   my $len1 = @$rows1;
   my $len2 = @$rows2;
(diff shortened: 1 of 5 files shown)


james-monkeyshines (approved):

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -278,16 +278,33 @@ sub row_subtotals {
...
-    die "Invalid SQL statement for subtotals. Must select a single column first, then a count.\n($sql1)";
-  }
-  if ($sql2 !~ /^\s*SELECT\s+[^,]+\s*,\s*COUNT[^,]+FROM.+GROUP\s+BY/ms) {
-    die "Invalid SQL statement for subtotals. Must select a single column first, then a count.\n($sql2)";
-  }
-
   my ( undef, $rows1 ) = _query( $dbc1, $sql1 );
   my ( undef, $rows2 ) = _query( $dbc2, $sql2 );
 
+  my $len1 = @$rows1;
```
james-monkeyshines, line 284: The t/TestDataCheck.t test is failing (see travis report https://travis-ci.org/github/Ensembl/ensembl-datacheck/builds/723484605), because it's using an incorrect SQL statement in order to generate an error which can then be checked, to see if it is raised appropriately.
But that 'incorrect' statement is returning a scalar not an arrayref, so this line is generating an untrapped error.
So, need to check here for an arrayref, and maybe add another test case to trap queries with more than two columns.

james-monkeyshines (changes requested):

james-monkeyshines (approved):

PR #311, opened 2020-09-28
Check for carriage returns in database varchar/text fields

james-monkeyshines:
Carriage returns in fields are problematic for dump files; easiest thing to do is remove them from the databases, and enforce continued absence with a datacheck. It's also useful to know about newlines or tabs in fields, but these do not present a particular problem for dumps, so checking those is done in a separate advisory datacheck.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/WhitespaceCritical.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/WhitespaceCritical.pm
@@ -1,0 +1,71 @@
+=head1 LICENSE
+
+Copyright [2018-2020] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/WhitespaceCritical.pm
```diff
@@ -0,0 +1,71 @@
...
+
+  foreach my $varchar (@$varchars) {
+    my ($table, $column) = @$varchar;
+    next if $table =~ /^MTMP/;
+
+    my $desc = "Column $table.$column contains no carriage returns";
+    my $diag = "Whitespace characters";
+    my $sql  = qq/
+      SELECT * FROM $table
+      WHERE $column REGEXP '\r'
```
marcoooo, line 65: Will the single quote not cause an issue there?

james-monkeyshines: No, I don't think so; single or double quotes should be fine for mysql.

marcoooo (approved):

PR #317, opened 2020-10-07
Added a check to test the UNIQUE KEYs

muffato:
See https://www.ebi.ac.uk/panda/jira/browse/ENSCORESW-3564 : UNIQUE KEYs that comprise a column that is NULL-able allow duplicated rows if they contain one NULL.
This is not something we want in the Compara schema, and it's caused some issues with the `theobroma_cacao` renaming this release, so here is a new DC to check those. It's structured the same way as ForeignKeysCompara and is quite generic. It could probably be applied onto other database types, should people need to.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/UniqueKeysCompara.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/UniqueKeysCompara.pm
@@ -1,0 +1,103 @@
+=head1 LICENSE
+
+Copyright [2018-2019] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::UniqueKeysCompara;
+
+use warnings;
+use strict;
+
+use Moose;
+use Path::Tiny;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use Bio::EnsEMBL::DataCheck::Utils qw/repo_location/;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'UniqueKeysCompara',
+  DESCRIPTION => 'Unique key relationships are not violated',
+  GROUPS      => ['compara', 'compara_gene_trees', 'compara_genome_alignments', 'compara_master', 'compara_syntenies'],
+  DB_TYPES    => ['compara'],
+  PER_DB      => 1
+};
+
+sub tests {
+  my ($self) = @_;
+
+  my $table_sql_file = $self->table_sql_file();
+
+  my @failed_to_parse;
+
+  my $table;
+  foreach my $line ( path($table_sql_file)->lines ) {
+    if ($line =~ /CREATE TABLE (IF NOT EXISTS )?`?(\w+)`?/) {
+      $table = $2;
+    } elsif (defined $table) {
+      next if $line =~ /^\-\-/;
+      next unless $line =~ /\bUNIQUE\b/;
+
+      $line =~ s/\(\d+\)//g;
+      my ($dum1, $dum2, $cols) = $line =~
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/UniqueKeysCompara.pm
```diff
@@ -0,0 +1,103 @@
...
+
+  my $table_sql_file = $self->table_sql_file();
+
+  my @failed_to_parse;
+
+  my $table;
+  foreach my $line ( path($table_sql_file)->lines ) {
+    if ($line =~ /CREATE TABLE (IF NOT EXISTS )?`?(\w+)`?/) {
+      $table = $2;
+    } elsif (defined $table) {
```
CristiGuijarro, line 51: I wondered if it might be handy to have a method that extracts the PRIMARY/FOREIGN/UNIQUE etc. keys that could be reused. Could also be incorporated into the ForeignKeys* tests:
```
SELECT 
    stat.table_name,
    GROUP_CONCAT(stat.column_name
        ORDER BY stat.seq_in_index
        SEPARATOR ', ') AS columns,
    tco.constraint_type
FROM
    information_schema.statistics stat
        JOIN
    information_schema.table_constraints tco ON stat.table_schema = tco.table_schema
        AND stat.table_name = tco.table_name
        AND stat.index_name = tco.constraint_name
WHERE
    stat.non_unique = 0
        AND stat.table_schema IN (SELECT DATABASE())
GROUP BY stat.table_schema , stat.table_name , stat.index_name , tco.constraint_type
ORDER BY stat.table_schema , stat.table_name;
```

muffato: Primary and unique keys can be retrieved this way, but not foreign keys because they're not recorded in MyISAM tables.
Also there is the risk that if somehow the unique key is dropped at some point (remember the issue we had with MySQLTransfer - it was erasing parts of the schema), we wouldn't be checking this table

CristiGuijarro: I think I was too specific with the example there! But what I meant more, was a method that generally (say in this case taking the `table.sql` file) pre-parses and extracts the assets for each table (as you have, but more generally). For example, returns per table objects with the keys, the engine, the columns etc, that could be applied to various other situations. In this case, the parsing of the table is very similar to that used in the `fk()` method. And then you would need to just collect from each object the necessary asset for the test. There is nothing wrong of course with the code, I was just thinking about clean recycling :slightly_smiling_face:

CristiGuijarro (approved):

lib/Bio/EnsEMBL/DataCheck/Checks/UniqueKeysCompara.pm
muffato: Ah, I see. Absolutely. I have copied and adapted ForeignKeysCompara, but I now realise that there is some similar code in [CompareSchema.pm](https://github.com/Ensembl/ensembl-datacheck/blob/release/103/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm). That's 4 files that would benefit from a bit of refactoring.

By the way, @james-monkeyshines, why is CompareSchema.pm not used on Compara databases ?

james-monkeyshines (approved):
This looks fine, and I take you point about having methods to parse SQL. However, was there a reason why you didn't query the mysql schema to get the unique keys? I think that'd tend to be a bit more robust, e.g. 
```
select
  table_name,
  index_name,
  group_concat(column_name order by seq_in_index) as index_columns,
  index_type
from
  information_schema.statistics
where
  table_schema = database() and
  non_unique = 0 and
  index_name <> 'PRIMARY'
group by table_name, index_name, index_type
order by index_name;
```

lib/Bio/EnsEMBL/DataCheck/Checks/UniqueKeysCompara.pm
james-monkeyshines: Compara dbs aren't included in the the CompareSchema datacheck because the format of the table.sql file is different: it contains the foreign keys (whereas other teams have them in a separate file). That was the reason they weren't included originally, but it's not really a good reason for that situation to persist - I should be able to update the code fairly easily so that it works for all dbs, it'd be reassuring to have it.

muffato: Thanks ! (for reference, a ticket has been created: ENSPROD-6233)

muffato:
Yes, good point, I could have queried the list of unique keys from the schema itself. I think I was worried the database wouldn't have the correct schema in some situations. We had an issue last release with MySQLTransfer because it copies the data _and_ the schema, and our source database was on an older version of the schema, so I preferred taking `table.sql` as the truth.

PR #320, opened 2020-11-04
Production DB-specific foreign key datachecks

james-monkeyshines:
Running the complete foreign key check in the Production DB Sync pipeline was taking way too long, and was largely redundant, so introduce new datachecks which only test the foreign keys that are affected by the sync. This involved a little bit of refactoring of the code for parsing the relationships from a file, in order to prevent code duplication. Also tweaked group membership a little, so that the pipeline can refer to the 'controlled_tables' group.


james-monkeyshines:
@muffato, @CristiGuijarro, FYI I pulled the foreign-key-from-sql-file code into a utility method (because I wanted to create a new datacheck that did a subset of FK tests), which has made the ForeignKeysCompara a bit neater.

vinay-ebi (approved):

PR #321, opened 2020-11-05
Updating tests of datacheck infrastructure

james-monkeyshines:
Half a dozen updates to different aspects of testing the datacheck framework - see individual commit messages for more details.


lib/Bio/EnsEMBL/DataCheck/DbCheck.pm
```diff
@@ -372,66 +372,26 @@ sub get_old_dba {
...
   } elsif ($group =~ /compara|ontology/) {
-    my $dbname = $self->dba->dbc->dbname;
-    my $current = $mca->schema_version;
-    $dbname =~ s/$current/$db_version/; # Ensembl version
-    my ($eg_current, $eg_version) = ($current-53, $db_version-53);
-    $dbname =~ s/$eg_current/$eg_version/; # EG version
-    $params{'-DBNAME'} = $dbname;
-    my $message = 'Previous version of database does not exist';
-    $dbh = $self->test_db_connection($uri, $params{'-DBNAME'}, $message);
+    ($params{'-DBNAME'}, $dbh) = $self->old_dbname_from_version(
```
marcoooo, line 375: Is it sensible to group the testing from `old_dbname_from_version` to code in `old_dbname_from_metadata`
Because the latter do a test on group as well. Hence reducing complexity of the if then else ?

lib/Bio/EnsEMBL/DataCheck/DbCheck.pm
```diff
@@ -372,66 +372,26 @@ sub get_old_dba {
...
+      $db_version,
+      $uri
+    );
   }
 
   # $old_dba can be undefined if there is no entry in the metadata db;
   # a datacheck could use the undefined-ness to skip tests in this case.
   my $old_dba;
-  if (exists $params{'-DBNAME'}) {
+  if (defined $params{'-DBNAME'}) {
```
marcoooo, line 394: What's the difference between 
`if (exists $params{'-DBNAME'}) {...`
and this statement? 
I mean: the code has two equivalent test done on same params and do some things from it. What's the point of having twice separated blocks?

lib/Bio/EnsEMBL/DataCheck/DbCheck.pm
```diff
@@ -473,6 +433,74 @@ sub get_old_dba {
   return $old_dba;
 }
 
+sub old_dbname_from_metadata {
```
marcoooo, line 436: This one would add the group test done before calling the other.

lib/Bio/EnsEMBL/DataCheck/DbCheck.pm
```diff
@@ -473,6 +433,74 @@ sub get_old_dba {
...
+    my $message = 'Database in metadata database does not exist';
+    $dbh = $self->test_db_connection($uri, $dbname, $message);
+  } elsif (scalar(@dbnames) > 1) {
+    die "Multiple release $db_version $group databases for $species";
+  }
+
+  return ($dbname, $dbh);
+}
+
+sub old_dbname_from_version {
```
marcoooo, line 490: This code would be integrated in the first one.

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
```diff
@@ -276,7 +276,6 @@ sub row_subtotals {
 
   $dbc2 = $dbc1 if ! defined $dbc2;
   $sql2 = $sql1 if ! defined $sql2;
-  $min_proportion = 1 if ! defined $min_proportion;
```
marcoooo, line 279: Was it just unused portion of code?

t/DbCheck_old.t
```diff
@@ -18,36 +18,176 @@ use warnings;
...
+      );
+      throws_ok(
+        sub { $check->get_old_dba },
+        qr/Old server details must be set/,
+        'Fail if old_server_uri is not set');
+
+      $check = TestChecks::DbCheck_1->new(
+        dba            => $dba,
+        server_uri     => $server_uri,
+        old_server_uri => $server_uri.'rhubarb_and_custard',
```
marcoooo, line 77: I don't like rhubarb, custard is ok though :-)

t/DbCheck_old.t
```diff
@@ -18,36 +18,176 @@ use warnings;
...
+      throws_ok(
+        sub { $check->get_old_dba },
+        qr/Database in metadata database does not exist/,
+        'Fail if database from metadata database does not exist (1/2)');
+      throws_ok(
+        sub { $check->get_old_dba($species, $db_type) },
+        qr/Database in metadata database does not exist/,
+        'Fail if database from metadata database does not exist (2/2)');
+
+      $old_dba = $check->get_old_dba('dinanthropoides_nivalis', $db_type);
```
marcoooo, line 111: :-D

marcoooo (changes requested):
Please can you think about merging the two new subs if possible?

lib/Bio/EnsEMBL/DataCheck/DbCheck.pm
james-monkeyshines: Agreed, change committed.

lib/Bio/EnsEMBL/DataCheck/DbCheck.pm
james-monkeyshines: The `exists` is checking if the key exists in the hash; the `defined` is checking the value corresponding to that key.

lib/Bio/EnsEMBL/DataCheck/Test/DataCheck.pm
james-monkeyshines: This was a bug - the behaviour should be the same as that of `row_totals`, where if the value is undefined, we seek an exact match. This line meant that the value was always defined, and it was impossible toy use the function to test for exact matches.

lib/Bio/EnsEMBL/DataCheck/DbCheck.pm
marcoooo: Then `defined` test if a value, `exists` the key exists, is it not quite the same True result?

marcoooo (approved):

PR #323, opened 2020-11-11
Catch 'flock' errors when opening config file

james-monkeyshines:
In the event of a locking error when opening a config file, try again after a 2 second delay.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Manager.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Manager.pm
@@ -167,7 +167,21 @@ sub load_config {
   if (defined $self->config_file) {
     die "Config file does not exist" unless -e $self->config_file;
 
-    my $json = path($self->config_file)->slurp;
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Manager.pm
```diff
@@ -167,7 +167,21 @@ sub load_config {
...
+    # it a second go if the first attempt fails.
+    my $json;
+    eval {
+      sleep(2);
+      $json = path($self->config_file)->slurp;
+    };
+    if ($@) {
+      sleep(2);
+      $json = path($self->config_file)->slurp;
+    }
```
radome, line 183: I think randomizing the sleep would help mitigating the problem, something like `sleep(int(rand(10)))` maybe?

james-monkeyshines: Yes, that's an excellent idea.

radome (approved):

PR #324, opened 2020-11-11
Check compara database schema

james-monkeyshines:
Include parsing of foreign keys in SQL schema files, to enable compara database schemas to be checked.


radome (approved):

lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
```diff
@@ -120,6 +123,10 @@ sub normalise_table_def {
   # at random. It controls padding with zeroes, and can safely be ignored.
   $table =~ s/INT\(\d+\)/INT/gm;
 
+  # The precision of FLOAT columns does not seem to be apparent from
+  # the table definitions, so ignore it to prevent spurious differences.
+  $table =~ s/FLOAT\(\d+\)/FLOAT/gm;
```
muffato, line 128: We have `FLOAT(5,2)` too, and `DOUBLE(5,4)` as well.
We've been talking about removing these numbers for integers (ENSCOMPARASW-2787) and we might eventually do it for floating-point values too, but that's not yet scheduled, and the priority is not very high

james-monkeyshines: The columns with two digits all match up between the sql and the current crop of databases, so didn't need to account for those (and in any case couldn't ignore them, because they force rounding, i.e. can actually affect the data stored). Providing a single float precision is only significant if the value is high enough (and it's <24 for everything in table.sql), and then it only affects the size of the storage.

PR #325, opened 2020-11-12
Exclude the repeatmask_repeatmodeler logic_name from the DC 

leannehaggerty:
Repeatmodeler is not currently used to mask the genome for subsequent analyses and should not be used to mask the genome for the ftp dumps (discussed in email) - therefore, the key for "repeatmask_repeatmodeler" is not needed in the meta table - skip it in the DC check.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
@@ -194,7 +194,7 @@ sub repeat_analysis {
 
     skip 'No repeat features', 1 unless scalar(@logic_names);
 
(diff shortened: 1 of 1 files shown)


james-monkeyshines (commented):
This is fine, but if I understand the use case correctly, I don't think this goes far enough - it's not enough to ignore the logic_name, you would want the datacheck to fail if it existed (assuming that non-verts agree and we don't need division-specific exceptions - will ask about that in the release meeting).

james-monkeyshines (approved):

james-monkeyshines:
Ignore my previous comment, I forgot how the test worked, and the proposed change is suitably comprehensive. I've just changed the test description slightly though, otherwise it could be a bit confusing.
Further tweaks may be necessary, depending on how non-vert divisions want to handle simlar situations, but will merge now so as not to block handovers.

PR #333, opened 2020-12-08
Update MetaKeyFormat url tests

james-monkeyshines:
Tests of url meta_keys need to tolerate special case of 'www.ensembl.org' and empty strings.


marcoooo (commented):
You sure that the value should always be `ensembl` ?

marcoooo (approved):

marcoooo:
>  'www.ensembl.org' and empty strings.



> You sure that the value should always be `ensembl` ?

Forget it, didn't see the `|` :-)

PR #336, opened 2020-12-16
Added Refseq peptide prefix RW 

vinay-ebi:


james-monkeyshines (commented):
'RW' is not a valid RefSeq prefix - it's a GenBank identifier prefix. What's the reason for trying to add this?

marcoooo:
List is fine, Related External DB has been added in production DB

PR #337, opened 2020-12-17
ExonBounds fix for circular chromosomes

james-monkeyshines:
Datacheck was relying on the assumption of linear regions, in order to make the queries a bit simpler. But it was failing to handle origin-spanning genes in bacterial circular regions; so need to be a bit more long-winded, to account for those. Introduce four separate queries, one for transcript start and one for transcript end, for each strand.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/ExonBounds.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/ExonBounds.pm
@@ -152,7 +152,7 @@ sub tests {
 
     if (defined $last_transcript_id && $last_transcript_id == $transcript_id) {
       if ($strand == 1) {
-        if ($last_end > $start && $last_start < $start) {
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/ExonBounds.pm
```diff
@@ -117,11 +152,11 @@ sub tests {
 
     if (defined $last_transcript_id && $last_transcript_id == $transcript_id) {
       if ($strand == 1) {
-        if ($last_end > $start) {
+        if ($last_end > $start && $last_start < $start) {
```
radome, line 155: Can I suggest rewriting this check using only `<` so the condition `last_start < start < last_end` is clearer
```suggestion
        if ($last_start < $start && $start < $last_end) {
```

radome (approved):

PR #339, opened 2021-01-14
add MD5 so that string comparison is case sensitive

brunocontrerasmoreira:
This patch makes core critical check MetaKeyAssembly case-sensitive @james-monkeyshines 

This was tested on pl1 with musa_acuminata_core_51_104_2 while fixing https://github.com/Ensembl/staging-patches/pull/443

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyAssembly.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyAssembly.pm
@@ -53,7 +53,7 @@ sub tests {
       cs.attrib RLIKE 'default_version' AND
       at.code = 'toplevel' AND
       m.meta_key = 'assembly.default' AND
-      MD5(cs.version) <> MD5(m.meta_value) AND
+      BINARY(cs.version) <> BINARY(m.meta_value) AND
       m.species_id = $species_id AND
       cs.species_id = $species_id
     GROUP BY
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyAssembly.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyAssembly.pm
@@ -1,6 +1,6 @@
 =head1 LICENSE
 
-Copyright [2018-2020] EMBL-European Bioinformatics Institute
+Copyright [2018-2021] EMBL-European Bioinformatics Institute
 
 Licensed under the Apache License, Version 2.0 (the 'License');
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyAssembly.pm
```diff
@@ -53,7 +53,7 @@ sub tests {
       cs.attrib RLIKE 'default_version' AND
       at.code = 'toplevel' AND
       m.meta_key = 'assembly.default' AND
-      cs.version <> m.meta_value AND
+      MD5(cs.version) <> MD5(m.meta_value) AND
```
marcoooo, line 56: Just a question, how big are the table on which you want to compute this? Because applying any function imply a MySQL full table-scan.

marcoooo: I tested it quickly on mentioned db, and moving to binary produce a gain of ~30%, not sure this is very important.

brunocontrerasmoreira: Hi @marcoooo , the meta and coord_system tables are small in my experience. But if you or @james-monkeyshines know of cheaper ways of comparing two strings with case-sensitivity I can try them, sure.

marcoooo: I was just thinking of using BINARY comparison instead of MD5. I actually don't know which one is faster. the only thing is that BINARY is a KEYWORD, and MD5 a mysql function. There stops my knowledge of both :-)

brunocontrerasmoreira: I will try BINARY, it seems like a cast. That should be faster than a function, but I need to make sure it works as expected

brunocontrerasmoreira: Quick test on pl1 with musa_acuminata_core_51_104_2:
i) MD5      39.4 ms
ii) BINARY 79.7 ms
Both seems to do the job

brunocontrerasmoreira:
Hi @marcoooo , the updated path now uses BINARY on the expectation that it will be less CPU hungry on average

marcoooo (approved):

marcoooo (changes requested):
BINARY is a key word, so the syntax should be without parenthesis. Is it working as such? I would have learned something new then!

marcoooo (approved):

brunocontrerasmoreira:
See https://www.geeksforgeeks.org/mysql-binary-function/

PR #341, opened 2021-02-02
Remove analysis db_version check from all groups

james-monkeyshines:
Check on analysis db_version is likely to fail for spurious reasons - too noisy, so remove from all groups for now. It is likely that this datacheck would only be useful for a subset of analyses, because there is little to be gained by updating databases to ensure consistency across the board.


james-monkeyshines:
Same as #340, but on the release branch.

marcoooo (approved):

PR #342, opened 2021-02-05
New dc cmp protein feature

at7:
Compare counts in protein_function_predictions and protein_function_predictions_attrib with counts from the previous release. Both tables get updated every release for human. The new dc will help spotting any missing data like we had for release 102 https://www.ebi.ac.uk/panda/jira/browse/ENSINT-647.


at7:
Could you please add @helensch as a reviewer?

helensch (approved):
Tested running on human 101 vs 102 and 102 vs 103.
Check works as expected

PR #344, opened 2021-02-18
Bugfixes/fix query check previous db

marcoooo:
In the context of the Rapid Release, we have multiple releases sharing the same ENS_VERSION number.
The query checking for previous release DB is therefore failing because for a given ens_version, all RR species may have same database name. 
Adding DISTINCT in the query should fix the issue, because 
- for main site, ens_version is still unique and we won't have the same use case, since db name would have change anyway.
- for RR, it ensure to return the only DB name as expected.


james-monkeyshines (approved):
This needs to be done in conjunction with an update to the config file, to explicitly set the release version of the dbs on the mirror server (otherwise the code will subtract 1 from the st5 version, which doesn't make sense for RR): mysql://ensro@mysql-ens-mirror-5:4672/102.
I've updated /nfs/panda/ensembl/production/datachecks/config/st5.json accordingly.

PR #345, opened 2021-02-18
Report bug fix from PR #344 on release 103

marcoooo:
Report back bug fix fro PR #344 onto release/103


james-monkeyshines (approved):
Doesn't matter a great deal, but not sure that this is necessary - the RR datachecks are using the release/104 branch.

PR #348, opened 2021-03-03
hotfix/repeatdetector_metakey

ndliberial:
Repeatdetector needs to be added to the exception list for repeat.analysis meta keys in same way as repeatmodeler is being skipped.


james-monkeyshines:
This isn't needed, since the same update has been made on the release/104 branch: https://github.com/Ensembl/ensembl-datacheck/pull/349
Note that Rapid Release uses the latest datacheck release branch, whatever the schema version; datachecks are essentially release-agnostic, so this is the easiest way of ensuring that the latest, bug-fixed, datachecks are applied.

PR #349, opened 2021-03-03
hot_fix/repeatdetector_metakey

ndliberial:
Added repeat detector to the list of logic names to be skipped under repeat.analysis meta keys.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
@@ -185,6 +185,8 @@ sub repeat_analysis {
         species_id = $species_id
       AND
         logic_name <> "repeatmask_repeatmodeler"
+      AND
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
```diff
@@ -185,6 +185,8 @@ sub repeat_analysis {
         species_id = $species_id
       AND
         logic_name <> "repeatmask_repeatmodeler"
+      AND
```
marcoooo, line 188: Personaly, I would have created a list, and use the list in the query instead of coupled "AND". Personal point of view.

james-monkeyshines: I agree, but this is clear enough, for now at least. If we get more exceptions in the future, a list would be neater.

james-monkeyshines (approved):

PR #351, opened 2021-03-04
New Compara Groups for RR

CristiGuijarro:
## Rapid Release in Compara requires a new group for two new database types.

- Most changes simply involve adding new group to existing checks.

- Some minor exceptions included into `ForeignKeysCompara` for both `compara_references` and `compara_homology_annotation`.

- Minor alteration in `CheckHomologyMLSS` to search for specific `ENSEMBL_HOMOLOGUES` `method.type` for `compara_homology_annotation`.

### Note: Basing against `master` branch because this is not release specific

PR #354, opened 2021-03-15
Bugfix incorrectly expressed regular expression

CristiGuijarro:
This is a little awkward - and I have no idea why it worked when I was initially testing (it does not work now).
See #351 - recent related PR.


marcoooo:
I would have seen these as equivalent :-)

marcoooo (approved):

PR #355, opened 2021-03-22
Advisory member_id check to integrate into compara pipelines

CristiGuijarro:
# Description

In the rapid release, datachecks will not be manually run on the compara reference updating pipeline. New SOP for the compara references in rapid release dictate that the `member_ids` are offset by the `genome_db_id` (to avoid overlaps in the other rapid_release pipelines which are run without a compara `master_db`) - this datacheck is for the purpose of ensuring this is the case for each `genome_db_id`

## Related Jira ticket:
[ENSCOMPARASW-4386](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4386)

### Change 1
The addition of a new Compara test method to generically test any offset_id ranges - we use these a lot and there are currently no DCs for these & perl unit testing additions

### Change 2
Implementing the above new test for the `seq_member_ids` in the `seq_member` table and `gene_member_ids` in the `gene_member` table


lib/Bio/EnsEMBL/DataCheck/Test/Compara.pm
```diff
@@ -145,4 +148,39 @@ sub cmp_tag {
...
+    SELECT COUNT(
+      DISTINCT(LEFT($table_column, $id_length))
+    )
+    FROM $table_name
+    WHERE $id_column = $id_column_val
+  /;
+
+  my $results = $helper->execute_single_result(-SQL => $sql);
+  my $test = "$table_column in $table_name is correctly offset by $id_column_val";
+  is( $results, 1, $test );
```
james-monkeyshines, line 183:
```suggestion
$tb->is_eq( $results, 1, $test );
```
... following on from last two comments: Rather than using the 'is' method directly, the idea is that we define an analogous method that involves an equality test.

lib/Bio/EnsEMBL/DataCheck/Test/Compara.pm
```diff
@@ -145,4 +148,39 @@ sub cmp_tag {
...
+  my $sql = qq/
+    SELECT COUNT(
+      DISTINCT(LEFT($table_column, $id_length))
+    )
+    FROM $table_name
+    WHERE $id_column = $id_column_val
+  /;
+
+  my $results = $helper->execute_single_result(-SQL => $sql);
+  my $test = "$table_column in $table_name is correctly offset by $id_column_val";
```
james-monkeyshines, line 182: The convention in the other custom test methods (and Perl tests generally) is to accept a `$name` parameter, which overrides anything the test generates. If that doesn't make sense in this case, fair enough, but thought it was worth mentioning.

CristiGuijarro: I have applied your suggestions and I think what you have said makes sense in this case too - so have included `$name` parameter and replaced `$test` with it. I will re-request your review to test I have understood :crossed_fingers:

lib/Bio/EnsEMBL/DataCheck/Test/Compara.pm
```diff
@@ -145,4 +146,43 @@ sub cmp_tag {
...
+out; if not provided, a default will be used.
+
+=back
+
+=cut
+
+sub check_id_range {
+  my ($dba, $table_name, $id_column, $id_column_val, $name) = @_;
+
+my $tb = $CLASS->builder;
```
JAlvarezJarreta, line 169:
```suggestion
  my $tb = $CLASS->builder;
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMemberIDRange.pm
```diff
@@ -0,0 +1,52 @@
...
+  TABLES         => ['gene_member', 'seq_member']
+};
+
+sub tests {
+  my ($self) = @_;
+  my $dba    = $self->dba;
+  my $helper = $dba->dbc->sql_helper;
+  my $genome_dbs   = $dba->get_GenomeDBAdaptor->fetch_all();
+
+  foreach my $genome_db ( sort { $a->dbID() <=> $b->dbID() } @$genome_dbs ) {
```
JAlvarezJarreta, line 45: Is there a reason behind sorting the genomes?

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMemberIDRange.pm
```diff
@@ -0,0 +1,52 @@
...
+    my $genome_db_id = $genome_db->dbID();
+    check_id_range($dba, "seq_member", "genome_db_id", $genome_db_id);
```
JAlvarezJarreta, line 47: `gene_member` table also needs to meet the same condition:
```suggestion
    check_id_range($dba, "gene_member", "genome_db_id", $genome_db_id);
    check_id_range($dba, "seq_member", "genome_db_id", $genome_db_id);
```

JAlvarezJarreta (changes requested):

lib/Bio/EnsEMBL/DataCheck/Checks/CheckMemberIDRange.pm
CristiGuijarro: It looks nice in the tap file when they are printed in order - also in terms of debugging later on, it could reveal obvious patterns for what may have gone wrong. But mostly because it looks nice in the tap output.

JAlvarezJarreta (approved):

james-monkeyshines (approved):

PR #356, opened 2021-03-29
Remove SchemaVersion from compara_references and update Compara schema

JAlvarezJarreta:
`SchemaVersion` was working in our previous `compara_references` pipeline because we were using release numbers in the same fashion as the release databases, but we have decided to create a master-like database instead (names `ensembl_compara_references`), so the release number as been removed and this DC is no longer applicable.

For the homology annotation pipeline, we have modified our schema to make things easier for us internally: since we use `InnoDB`, we are taking advantage of the `ON UPDATE CASCADE` option to automatically update some keys, so `ALTER TABLE` is now part of our schema. Thus, `Utils::foreign_keys()` method has been updated.

Both edits have been tested in the [latest run of our `UpdateReferencesDB` pipeline](http://guihive.ebi.ac.uk:8080/versions/96/?driver=mysql&username=ensro&host=mysql-ens-compara-prod-3&port=4523&dbname=jalvarez_update_references_103_fresh).

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SchemaVersion.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SchemaVersion.pm
@@ -29,7 +29,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME        => 'SchemaVersion',
   DESCRIPTION => 'The schema version meta key matches the DB name',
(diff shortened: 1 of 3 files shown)


JAlvarezJarreta:
I can update the compara `table.sql` if you want to make coveralls happy.

CristiGuijarro (approved):

james-monkeyshines (approved):
Looks fine, when we update our test dbs in due course, we should get the updated SQL file anyway.

PR #365, opened 2021-05-05
Changes to account for per-species compara databases

CristiGuijarro:
## Description

Additional changes for rapid release in Compara, in particular the per-species compara databases. 

- The homology MLSS may not be retrievable using the traditional compara API, so a more simple check is reverted to. This should be fine because the same data is already tested in the pipeline, where the API does work.
- The reference genomes are not retained in the per-species compara database, so a reduced `ForeignKeysCompara` check is necessary, although again, the data is checked in the pipeline first, so this should be enough.

## Related EPIC in compara

[ENSCOMPARASW-3545](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-3545)

## Testing

Tested in homology annotation pipeline database, main release database and per-species compara database


dthybert (approved):
I do approve it

james-monkeyshines (approved):

PR #369, opened 2021-05-12
Better configuration of available servers/databases

james-monkeyshines:
Make server_uri and old_server_uri multi-valued.

For old_server_uri, we simply loop through the list, looking for the old version of the relevant db, stopping when we find it.

For server_uri, it remains as an alternative to specifying a registry_file, and in this case we loop through all values, adding them to the registry. There is some additional functionality, however - if both registry_file and server_uri are given, then only the registry is loaded by default. But, if an ancillary database (e.g. a core db needed for a corelike datacheck) is required, we first loop through the server_uri values, stopping when we find it; if not found there, we look in the registry. This functionality allows us to have a generic all-purpose registry file, which we can override for specific use cases, for example: genebuild create an rnaseq db, and test it against a copy of the core db on the same server (they only need to specify a registry that covers that server); at handover, we add the target server at the front of the server_uri list, so that testing can be done aginst the core db there (and ignoring the genebuild core db since it may be a slightly different version).

See https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-6710 for more details on the use cases.

### DIFF
--- a/t/DbCheck.t
+++ b/t/DbCheck.t
@@ -372,7 +372,7 @@ subtest 'Registry instantiation', sub {
 
   $check = TestChecks::DbCheck_1->new(
     dba        => $dba,
-    server_uri => $server_uri,
+    server_uri => [$server_uri],
   );
 
   # The registry object is a bit weird, the return value is a string,
@@ -383,7 +383,7 @@ subtest 'Registry instantiation', sub {
 
   $check = TestChecks::DbCheck_1->new(
     dba        => $dba,
-    server_uri => $server_uri.$dba->dbc->dbname,
+    server_uri => [$server_uri.$dba->dbc->dbname],
   );
 
   throws_ok(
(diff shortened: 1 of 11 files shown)


t/DbCheck.t
```diff
@@ -372,7 +372,7 @@ subtest 'Registry instantiation', sub {
 
   $check = TestChecks::DbCheck_1->new(
     dba        => $dba,
-    server_uri => $server_uri,
+    server_uri => [$server_uri],
```
marcoooo, line 375: Is this turning initial input in an array? Should we consider renaming the variable to something plural? (it might be a bit dangerous though!)

marcoooo (commented):
I am heavily relying on your wisdom!

marcoooo (approved):

t/DbCheck.t
james-monkeyshines: Yes, it's now an array - but I figured changing the name was more effort than it was worth. The scripts for running datachecks/pipelines, which are where most users will interact with the parameter, still accept string values, i.e. you use `--server_uri mysql://etc`, as before. So from a user perspective, it'd probably be unnecessarily disruptive to change the name.

marcoooo: Fine, seeing the number of time this param is used, it's sensible, Thanks.

PR #371, opened 2021-05-17
Rename queue.

james-monkeyshines:


james-monkeyshines:
Will rebase against master once the feature/server_uri branch is merged, and then take out of draft status.

marcoooo (approved):
I believe we need the same in production repo as well.

james-monkeyshines:
> I believe we need the same in production repo as well.

Something like this, you mean? :wink: 
https://github.com/Ensembl/ensembl-production/compare/feature/merge_104...feature/codon_migration

PR #372, opened 2021-05-20
datacheck added to check meta key species.common name is simillar acr…

vinay-ebi:
All V species that form a group of strains or breeds need to have the same value for species.common_name; if they don't then they are not grouped together in Search. If the metadata DB knows which species are grouped in this way, we could have a DC to make sure the species.common_names match

Test case: 
perl scripts/run_datachecks.pl  $(${ST1} details script)   -config_file  datachecks/config/${ST1}.json -name SpeciesCommonName  -dbname mus_musculus_lpj_core_104_1  -r datachecks/registry/sta-1.pm -history_file history.json

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
@@ -23,7 +23,6 @@ use strict;
 
 use Moose;
 use Test::More;
-use Data::Dumper;
 
 extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 
@@ -74,4 +73,3 @@ sub CheckCommonName {
 }
 
 1;
-
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
@@ -70,7 +70,7 @@ sub check_common_name {
         	my $mca = $strain_dba->get_adaptor("MetaContainer");
         	my $dbname =  $genome->dbname();
                 my $common_name = $mca->single_value_by_key('species.common_name') ? $mca->single_value_by_key('species.common_name')  : "No meta_key species.common_name in $dbname";
-                my $desc =  "Meta key species.common_name is similar in DB $dbname for strain group $strain_group";
+                my $desc =  "Meta key species.common_name is the same in DB $dbname for strain group $strain_group";
                 is($species_common_name, $common_name, $desc)
        } 
   }    
@@ -79,4 +79,3 @@ sub check_common_name {
 
 
 1;
-
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
@@ -28,7 +28,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 
 use constant {
   NAME        => 'SpeciesCommonName',
-  DESCRIPTION => 'Meta key species.common_name should be same for species form a group of strains or breed',
+  DESCRIPTION => 'Meta key species.common_name should be same for species from a group of strains or breeds',
   GROUPS      => ['core', 'meta'],
   DB_TYPES    => ['core'],
   TABLES      => ['meta'],
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
@@ -69,7 +69,7 @@ sub check_common_name {
 
         	my $mca = $strain_dba->get_adaptor("MetaContainer");
         	my $dbname =  $genome->dbname();
-                my $common_name = $mca->single_value_by_key('species.common_name') ? $mca->single_value_by_key('species.common_name')  : "No meta_key species.common_name in $dbname";
+                my $common_name = $mca->single_value_by_key('species.common_name');
                 my $desc =  "Meta key species.common_name is the same in DB $dbname for strain group $strain_group";
                 is($species_common_name, $common_name, $desc)
        } 
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
@@ -53,7 +53,7 @@ sub tests {
 
 sub CheckCommonName {
   my ($self, $division, $reference_species_common_name, $strain_group) = @_;
-  my $gdba = $self->registry->get_DBAdaptor("multi", "metadata")->get_GenomeInfoAdaptor();
+  my $gdba = $self->get_dba("multi", "metadata")->get_GenomeInfoAdaptor();
   my %unique_common_name;
   for my $genome (@{$gdba->fetch_all_by_division($division)}) {
         if($genome->strain() and $genome->reference() and $genome->reference() eq $strain_group){
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
@@ -58,7 +58,7 @@ sub check_common_name {
 
   my ($self, $division, $species_common_name,  $strain_group) = @_;
   my $gdba = $self->get_dba("multi", "metadata")->get_GenomeInfoAdaptor();
-  my %unique_common_name;
+
   for my $genome (@{$gdba->fetch_all_by_division($division)}) {
         if($genome->reference() and $genome->reference() eq $strain_group){
         	my $strain_name = $genome->name;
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
+=head1 LICENSE
+
+Copyright [2018-2020] EMBL-European Bioinformatics Institute
```
marcoooo, line 3: Copyright to 2021 please.

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+}
+
+sub CheckCommonName {
+  my ($self, $division, $reference_species_common_name, $strain_group) = @_;
+  my $gdba = $self->registry->get_DBAdaptor("multi", "metadata")->get_GenomeInfoAdaptor();
+  my %unique_common_name;
+  for my $genome (@{$gdba->fetch_all_by_division($division)}) {
+        if($genome->strain() and $genome->reference() and $genome->reference() eq $strain_group){
```
marcoooo, line 60: Not sure about the second part here `and $genome->reference()` 
Aren't we supposed to test the strains and check that they hold the same as the reference?

marcoooo (changes requested):

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
vinay-ebi: sure i will change it.

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
vinay-ebi: $strain_group holds the required value to comapre with reference $genome->reference() eq $strain_group

marcoooo: :+1:

marcoooo (approved):

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+  DESCRIPTION => 'Meta key species.common_name should be same for group of strains or breed',
+  GROUPS      => ['core', 'meta'],
+  DB_TYPES    => ['core'],
+  TABLES      => ['meta'],
+};
+
+sub tests {
+  my ($self) = @_;
+  my $mca = $self->dba->get_adaptor("MetaContainer");
+  my $strain = $mca->single_value_by_key('species.strain');
```
james-monkeyshines, line 41: Don't need to get/check this value - the important one for this check is 'species.strain_group'.

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+  my $strain_group = $mca->single_value_by_key('species.strain_group'); 
+  my $reference_species_common_name = $mca->single_value_by_key('species.common_name');
```
james-monkeyshines, line 43: This isn't necessarily the reference species - this datacheck might be running on one of the strains.

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+        
+	$self->CheckCommonName($division, $reference_species_common_name, $strain_group);
+  }
+  else{
+	skip('No Strains or Breeds for Species');
+  }
+
+}
+
+sub CheckCommonName {
```
james-monkeyshines, line 55: Doesn't matter a great deal, but the convention for method names is snake case, rather than camel.

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+  my ($self, $division, $reference_species_common_name, $strain_group) = @_;
+  my $gdba = $self->registry->get_DBAdaptor("multi", "metadata")->get_GenomeInfoAdaptor();
+  my %unique_common_name;
+  for my $genome (@{$gdba->fetch_all_by_division($division)}) {
+        if($genome->strain() and $genome->reference() and $genome->reference() eq $strain_group){
```
james-monkeyshines, line 60: Could lose the `$genome->strain()`, by definition it wil be defined if `$genome->reference()` is.

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+        	my $mca = $self->registry->get_adaptor( $genome->name(), 'Core', 'MetaContainer' );
```
james-monkeyshines, line 61: For the reasons described above, better to use `get_dba`. Also, we probably have this db available in our registry, but it's not guaranteed. We can provide a more useful failure message if we check if it is defined.
```suggestion
        	my $strain_name = $genome->name;
        	my $strain_dba = $self->get_dba($strain_name, 'core');
        	my $desc_strain_dba = "Core database for $strain_name found";
        	my $pass = ok(defined $strain_dba, $desc_strain_dba);
        	next unless $pass;

        	my $mca = $strain_dba->get_adaptor("MetaContainer");
```

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+	$self->CheckCommonName($division, $reference_species_common_name, $strain_group);
+  }
+  else{
+	skip('No Strains or Breeds for Species');
+  }
+
+}
+
+sub CheckCommonName {
+  my ($self, $division, $reference_species_common_name, $strain_group) = @_;
```
james-monkeyshines, line 56: `$reference_species_common_name` is not used in this method. If it is to be used, probably better to call it `$species_common_name`

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+  my $gdba = $self->registry->get_DBAdaptor("multi", "metadata")->get_GenomeInfoAdaptor();
+  my %unique_common_name;
+  for my $genome (@{$gdba->fetch_all_by_division($division)}) {
+        if($genome->strain() and $genome->reference() and $genome->reference() eq $strain_group){
+        	my $mca = $self->registry->get_adaptor( $genome->name(), 'Core', 'MetaContainer' );
+        	my $dbname =  $genome->dbname();
+                my $common_name = $mca->single_value_by_key('species.common_name') ? $mca->single_value_by_key('species.common_name')  : "No meta_key species.common_name in $dbname";
+                push(@{$unique_common_name{$common_name}} , $dbname);
```
james-monkeyshines, line 64: This is checking all pairwise combinations from the set of common names across all the dbs - which is fine, but I don't think any other datachecks work that way; they all check the current db (given by `self->dba`) against a set of other dbs. So I think there needs to be a comment to make it clear that this is deviating from the norm.

(An alternative approach would be to check if `$species_common_name` and `$common_name` match, within the loop here.)

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+                push(@{$unique_common_name{$common_name}} , $dbname); 
+       } 
+  }    
+
+  my $report='';
+  for my $common_name (keys %unique_common_name){
+  	$report.= join("\n$common_name: ", @{$unique_common_name{$common_name}});
+  }
+  my $desc = " Meta key species.common_name is similar in all DBs for strain group $strain_group \n". $report;
+  is(keys %unique_common_name ,  1 , $desc) or diag('Meta key species.common name is  not similar in dbs : ' . $report);
```
james-monkeyshines, line 73: diag here is unnecessary - it is just repeating the error message. diags are useful when there is a failure, to provide additional information to help debugging, which would otherwise clog up the test results. Actually, should probably remove the `\n$report` from `$desc`, and do `diag($report)`.

james-monkeyshines (changes requested):

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
```diff
@@ -0,0 +1,77 @@
...
+  }
+  else{
+	skip('No Strains or Breeds for Species');
+  }
+
+}
+
+sub CheckCommonName {
+  my ($self, $division, $reference_species_common_name, $strain_group) = @_;
+  my $gdba = $self->registry->get_DBAdaptor("multi", "metadata")->get_GenomeInfoAdaptor();
```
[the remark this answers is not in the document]
vinay-ebi: sure will use  get_dba method

lib/Bio/EnsEMBL/DataCheck/Checks/SpeciesCommonName.pm
vinay-ebi: var name misleading will rename to species_common_name

james-monkeyshines (approved):
Couple of suggestions for changing the wording, code looks good.

james-monkeyshines:
@vinay-ebi  The index file needs to be updated to include the new datacheck.

PR #381, opened 2021-06-08
Hotfix/repeat nonvert

jamie-m-a:
Altered logic so that all non-verts can have repeatdetector.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
@@ -175,6 +175,16 @@ sub repeat_analysis {
     # in order to do a count, use SQL rather than API.
     my $helper = $self->dba->dbc->sql_helper;
     my $species_id = $self->dba->species_id;
+    my $mca = $self->dba->get_adaptor('MetaContainer');
+    my @rep_list = ();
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
```diff
@@ -177,11 +177,11 @@ sub repeat_analysis {
     my $species_id = $self->dba->species_id;
     my $mca = $self->dba->get_adaptor('MetaContainer');
     my @rep_list = ();
-    if ($mca->get_division eq 'EnsemblPlants') {
-      @rep_list = qw("repeatmask_repeatmodeler");
+    if ($mca->get_division eq 'EnsemblVertebrates') {
+      @rep_list = qw("repeatmask_repeatmodeler" "repeatdetector);
```
marcoooo, line 181: IS it not missing some kind of `"` char here?

marcoooo (changes requested):

lib/Bio/EnsEMBL/DataCheck/Checks/MetaKeyConditional.pm
jamie-m-a: Yes, it might be! Added.

marcoooo (approved):

marcoooo (approved):

PR #386, opened 2021-06-29
ControlledTablesCompara taxonomy compatibility with compara pipelines

CristiGuijarro:
At a loss as to why this works and the other does not. For whatever reason, the previous way was incompatible with use in pipeline via `RunDataChecks.pm` and compara style registry, the taxonomy was undetectable. This has now been tested in both compara pipeline and manual datacheck commands, and appears to work. Needs to be merged ASAP to prevent blocking compara production pipelines.
[ENSCOMPARASW-4649](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4649)


james-monkeyshines (approved):
The registry is a riddle wrapped in a mystery inside an enigma...

PR #388, opened 2021-07-01
File dump-related datacheck updates. 

james-monkeyshines:
Added a check for Rapid Release display name.

Split the existing check on INSDC names into two, because only part of it is needed for exporting GFF3 formatted for ENA submission, and we want to be able to run just that bit. The INDSC name check requires all top-level sequences to have an INSDC name/synonym (rather than a set of named coord_systems), which matches the ENA submission use case.

Both of these datachecks are intended to be run within file dump pipelines, to check that the core databases contain all the information required for dumping (https://github.com/Ensembl/ensembl-production/pull/576).

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/index.json
+++ b/lib/Bio/EnsEMBL/DataCheck/index.json
@@ -2170,12 +2170,23 @@
       "name" : "SeqRegionNames",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::SeqRegionNames"
    },
+   "SeqRegionNamesENA" : {
+      "datacheck_type" : "advisory",
+      "description" : "Seq_region contig names from ENA are appropriately formatted and attributed",
+      "groups" : [
+         "assembly",
+         "core"
+      ],
+      "name" : "SeqRegionNamesENA",
+      "package_name" : "Bio::EnsEMBL::DataCheck::Checks::SeqRegionNamesENA"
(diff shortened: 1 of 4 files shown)


lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -1113,6 +1113,15 @@
       "name" : "DensitySNPs",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::DensitySNPs"
    },
+   "DisplayNameFormat" : {
+      "datacheck_type" : "critical",
+      "description" : "For Rapid Release, the display name must be a specific format",
+      "groups" : [
+         "rapid_release"
```
marcoooo, line 1120: How would rapid_release set to be run only for RR handover?

james-monkeyshines: The datacheck execution part of the handover service now has a concept of the target server, with the 'staging_uri' parameter (https://github.com/Ensembl/ensembl-prodinf-handover/blob/main/src/ensembl/production/handover/celery_app/utils.py#L224). So could use that to add the 'rapid_release' group to the set of datachecks that are executed.

marcoooo: Awesome!

marcoooo (approved):

PR #389, opened 2021-07-08
Check IDs are in correct range for associated MLSS ID

awgymer:
## Description

 We try to ensure IDs (i.e. `genomic_align_id`) are within the range `[mlss_id * 10 ^ 10 .. (mlss_id + 1) * 10 ^ 10]` for traceability. It was detected that some older IDs had not been set to adhere to this and although it's not a breaking issue it is good to flag any cases where this is not the case. This should only affect release DBs.  

This DC will check the following ID columns:

- `genomic_align.genomic_align_id`
- `genomic_align_block.genomic_align_block_id`
- `genomic_align_block.group_id`
- `genomic_align_tree.node_id`
- `constrained_element.constrained_element_id`
- `dnafrag.dnafrag_id`
- `conservation_score.genomic_align_block_id` 

**Related JIRA tickets:**
- ENSCOMPARASW-4632

## Testing
Run against the (now fixed) `ensembl_compara_51_104` DB and passed. 
Before adding skip for pipeline DBs (which would and *should* fail) ran against `agymer_wheat_lastz_batch18` (`cp9`) and returned errors for `group_id` as expected.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
@@ -59,28 +59,38 @@ sub tests {
   is_rows_zero($self->dba, $sql_1, $desc_1);
 
 
-  my $desc_2 = "genomic_align_block.genomic_align_block_id all within correct MLSS range";
+  my $desc_2 = "genomic_align.genomic_align_block_id all within correct MLSS range";
   my $sql_2 = q/
     SELECT genomic_align_block_id, method_link_species_set_id
-    FROM genomic_align_block 
+    FROM genomic_align
     WHERE genomic_align_block_id < (method_link_species_set_id * 10000000000) 
       OR genomic_align_block_id > ((method_link_species_set_id + 1) * 10000000000)
   /;
   is_rows_zero($self->dba, $sql_2, $desc_2);
 
 
-  my $desc_3 = "genomic_align_block.group_id all within correct MLSS range";
+  my $desc_3 = "genomic_align_block.genomic_align_block_id all within correct MLSS range";
   my $sql_3 = q/
+    SELECT genomic_align_block_id, method_link_species_set_id
+    FROM genomic_align_block 
+    WHERE genomic_align_block_id < (method_link_species_set_id * 10000000000) 
+      OR genomic_align_block_id > ((method_link_species_set_id + 1) * 10000000000)
+  /;
+  is_rows_zero($self->dba, $sql_3, $desc_3);
+
+
+  my $desc_4 = "genomic_align_block.group_id all within correct MLSS range";
+  my $sql_4 = q/
     SELECT group_id, method_link_species_set_id
     FROM genomic_align_block 
     WHERE group_id < (method_link_species_set_id * 10000000000) 
       OR group_id > ((method_link_species_set_id + 1) * 10000000000)
   /;
-  is_rows_zero($self->dba, $sql_3, $desc_3);
+  is_rows_zero($self->dba, $sql_4, $desc_4);
 
 
-  my $desc_4 = "genomic_align_tree.node_id all within correct MLSS range";
-  my $sql_4 = q/
+  my $desc_5 = "genomic_align_tree.node_id all within correct MLSS range";
+  my $sql_5 = q/
     SELECT DISTINCT(t.node_id), g.method_link_species_set_id 
     FROM genomic_align_tree t 
     JOIN genomic_align g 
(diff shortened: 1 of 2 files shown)


james-monkeyshines (approved):

lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
```diff
@@ -0,0 +1,132 @@
...
+  GROUPS         => ['compara', 'compara_pairwise_alignments'],
+  DATACHECK_TYPE => 'advisory',
+  DB_TYPES       => ['compara'],
+  TABLES         => ['genomic_align', 'genomic_align_block', 'genomic_align_tree', 'constrained_element', 'conservation_score', 'dnafrag']
+};
+
+
+sub skip_tests {
+  my ($self) = @_;
+  if (is_compara_ehive_db($self->dba) != 1) {
```
JAlvarezJarreta, line 43: You want to skip the non-release databases, right?
```suggestion
  if (is_compara_ehive_db($self->dba) == 1) {
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
```diff
@@ -0,0 +1,132 @@
...
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+use Bio::EnsEMBL::DataCheck::Utils qw( is_compara_ehive_db );
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME           => 'CheckIDRangeMLSS',
+  DESCRIPTION    => 'All relevant IDs are within the offset range for their MLSS ID',
+  GROUPS         => ['compara', 'compara_pairwise_alignments'],
```
JAlvarezJarreta, line 34: This DC also applies to MSAs, that run GERP and compute genomic align trees:
```suggestion
  GROUPS         => ['compara', 'compara_pairwise_alignments', 'compara_genome_alignments'],
```

lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
```diff
@@ -0,0 +1,132 @@
...
+  my $desc_7 = "conservation_score.genomic_align_block_id all within correct MLSS range";
+  my $sql_7 = q/
+    SELECT cs.genomic_align_block_id, g.method_link_species_set_id
+    FROM conservation_score cs 
+    JOIN genomic_align_block g 
+      ON g.genomic_align_block_id=cs.genomic_align_block_id
+    WHERE g.genomic_align_block_id < (g.method_link_species_set_id * 10000000000) 
+      OR g.genomic_align_block_id > ((g.method_link_species_set_id + 1) * 10000000000)
+  /;
+  is_rows_zero($self->dba, $sql_7, $desc_7);
```
JAlvarezJarreta, line 128: `CheckConservationScore.pm` (critical) already checks that all `conservation_score.genomic_align_block_id` match one `genomic_align_block.genomic_align_block_id`, and you have previously checked that `genomic_align_block.genomic_align_block_id` are within the correct range, so I think this check can be skipped:
```suggestion
```

lib/Bio/EnsEMBL/DataCheck/index.json
```diff
@@ -403,6 +403,16 @@
       "name" : "CheckHomologyMLSS",
       "package_name" : "Bio::EnsEMBL::DataCheck::Checks::CheckHomologyMLSS"
    },
+   "CheckIDRangeMLSS" : {
+      "datacheck_type" : "advisory",
+      "description" : "All relevant IDs are within the offset range for their MLSS ID",
+      "groups" : [
+         "compara",
+         "compara_pairwise_alignments"
```
JAlvarezJarreta, line 411: Following my previous comment:
```suggestion
         "compara_pairwise_alignments",
         "compara_genome_alignments"
```

JAlvarezJarreta (changes requested):

lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
awgymer: I'd actually made this change and just forgotten to commit it before opening the PR

lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
awgymer: I thought when we discussed this midweek it was decided it made sense to report all tables where changes would be required? Yes, the other check would flag a problem, but this DC wouldn't make it clear you need to ensure consistency with `conservation_score` was my understanding?

JAlvarezJarreta: Sorry if there was a misunderstanding, what I meant is that we want to report all tables affected, but if something is wrong in `conservation_score` and `CheckConservationScore.pm` has passed, then the source of the issue is in `genomic_align_block`. Thus, following the database schema should lead the person to fix it here as well. Furthermore, if you would prefer to leave this, then you need to add a check for `genomic_align.genomic_align_block_id` as well. I will leave it up to you which one of the two options you prefer :)

lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
```diff
@@ -0,0 +1,131 @@
...
+  my $sql_6 = q/
+    SELECT d.dnafrag_id, g.method_link_species_set_id 
+    FROM dnafrag d 
+    JOIN genomic_align g 
+      ON g.dnafrag_id=d.dnafrag_id
+    WHERE d.name LIKE 'Ancestor%'
+      AND (
+        d.dnafrag_id < (g.method_link_species_set_id * 10000000000) 
+        OR d.dnafrag_id > ((g.method_link_species_set_id + 1) * 10000000000)
+      )
```
JAlvarezJarreta, line 114: Just as a curiosity, if you wanted to avoid the `JOIN` (I have tested and it is not that slow, so it is not worth to change it, I think), you could have used `SUBSTRING_INDEX(SUBSTRING_INDEX(d.name, '_', 2), '_', -1)` to get the MLSS ID from the `dnafrag` table.

awgymer: I assumed the `name` is datachecked elsewhere,  but where possible I always think it is best to go to the source when comparing data such as IDs.

lib/Bio/EnsEMBL/DataCheck/Checks/CheckIDRangeMLSS.pm
awgymer: Hmm I think prefer to check all possible cases for consistency so I will add the extra test.

JAlvarezJarreta (approved):

PR #393, opened 2021-07-12
EP-6798: Skip some Compare DCs if assembly or geneset is new

luca-drf:


lib/Bio/EnsEMBL/DataCheck/Utils.pm
```diff
@@ -321,4 +331,56 @@ sub is_compara_ehive_db {
...
+Takes two MetaContainer adaptors (C<$mca> and C<$old_mca>) and returns 1 if the
+two databases have the same assembly, 0 otherwise.
+
+=back
+
+=cut
+sub same_assembly {
+  my ($mca, $old_mca) = @_;
+
+  return same_metavalue($mca, $old_mca, 'asembly.default');
```
james-monkeyshines, line 367:
```suggestion
  return same_metavalue($mca, $old_mca, 'assembly.default');
```
This typo didn't generate any obvious errors, because if you ask for a meta_key that doesn't exist, the Perl API gives you an 'undef' value, rather than failing. I think the behaviour we want would be for same_metavalue to be false if either value is undefined, since in the absence of information, it seems safest to assume difference?

james-monkeyshines (changes requested):

lib/Bio/EnsEMBL/DataCheck/Utils.pm
luca-drf: Yep, thank you, well spotted!

james-monkeyshines (approved):

PR #396, opened 2021-07-28
Fix: should fail if not all translations have attribs

MatBarba:
Change the min_proportion to undef, because we need the number of
translations to be the number of attribs of each type. The value 1
basically meant that any number of attribs would work, as long as it
wasn't more than the number of translations.


luca-drf:
Thank you for fixing that @MatBarba, would you mind cherry picking ae05ab0 onto `release/105` and open a new PR, please?

MatBarba:
Here it is for 105: https://github.com/Ensembl/ensembl-datacheck/pull/397

PR #399, opened 2021-08-19
Allow for differing clustersets/additional trees in homology relationships

CristiGuijarro:
### CheckHomology

:bangbang: | This needs to be merged to handover compara vertebrates
:---: | :---

This compara DC checks for redundancy in homology relationships - duplicate relationships for a pair of `gene_members`.
But it does not take into account the possibility that a homology relationship can be identical and inferred from two different trees.
This flexibility has been checked on the staging sites and in the compara API, and causes no issues.

* Non default clustersets with overlapping mlsss in
  in the strains/breeds and default can cause `gene_members`
  to have duplicated relationships in different trees

* These "redundant" or duplicate relationships are actually
  a sign of consistency between trees

* Whilst overlapping mlsss should be avoided in the future,
  the data is not at fault, not technically redundant and
  should be allowed

* Check now groups by gene_tree_root_id as well as
  gene_members

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckHomology.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckHomology.pm
@@ -44,16 +44,22 @@ sub tests {
   is_one_to_many($dbc, "homology_member", "homology_id", $desc_1);
   ### Hoping for a better idea than this query (below)
   my $hideous_sql = q/
-    SELECT hm1.gene_member_id gene_member_id1, hm2.gene_member_id gene_member_id2, COUNT(*) num, 
-      GROUP_CONCAT(h1.description order by h1.description) descs 
-        FROM homology h1 
-          CROSS JOIN homology_member hm1 
-            USING (homology_id)
-          CROSS JOIN homology_member hm2 
-            USING (homology_id)
-    WHERE hm1.gene_member_id < hm2.gene_member_id
-      GROUP BY hm1.gene_member_id, hm2.gene_member_id 
-        HAVING COUNT(*) > 1
+    SELECT
+      hm1.gene_member_id gene_member_id1,
+      hm2.gene_member_id gene_member_id2,
+      COUNT(*) num,
+      GROUP_CONCAT(h1.description
+          ORDER BY h1.description) descs
+    FROM
+      homology h1
+          CROSS JOIN
+      homology_member hm1 USING (homology_id)
+          CROSS JOIN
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CheckHomology.pm
```diff
@@ -44,16 +44,22 @@ sub tests {
...
+    SELECT
+      hm1.gene_member_id gene_member_id1,
+      hm2.gene_member_id gene_member_id2,
+      COUNT(*) num,
+      GROUP_CONCAT(h1.description
+          ORDER BY h1.description) descs
+    FROM
+      homology h1
+          CROSS JOIN
+      homology_member hm1 USING (homology_id)
```
marcoooo, line 56: imho, since `CROSS JOIN` is filtered on `homology_id` it will produce the exact same result as an inner join, I am not sure if that has an impact on performance, (depend on how mysql optimize the resulting data access) but CROSS JOIN is generally a "I want to kill my server" kind of thing.

CristiGuijarro: This is a horrible "legacy" style query that was simply taken from the java healthcheck version, and now altered. I honestly strongly dislike the entire query, but have been unable to improve upon it. We do need to get all the possible lines in the table because of the nature of the way the relationships are defined in the `homology` and `homology_member` tables - really the best we can hope for is an improved schema design/model to store these homology relationships :crossed_fingers:

marcoooo: As I said the CROSS being filtered on `homology_id` the result is NOT all the possible lines though, only the matched ones. I was just concerned about MySQL doing the CROSS `then` filtering, or automatically move to a more simple `INNER`.

marcoooo: I'll wait for @twalsh-ebi approval before merging anyway

CristiGuijarro: @marcoooo - Thank you and I think what you say makes sense. I will write a ticket for improving the query on our side based on what you have said for another release, at the moment I don't think we have the time to test another change

marcoooo: NP.

marcoooo (approved):

PR #401, opened 2021-09-03
Check for mitochondrial attribute

james-monkeyshines:
We expect MT seq_regions to have a 'sequence_location' attrib, with a value of 'mitochondrial_chromosome'. This was only being checked if the coord_system name was 'chromosome' - but most new core dbs do not use that name, they have 'primary_assembly' instead.

This update splits out the MT stuff to a new datacheck, so that it does not depend on the coord_system name.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/ChromosomesAnnotated.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/ChromosomesAnnotated.pm
@@ -59,12 +59,6 @@ sub tests {
       my $sr_name = $_->seq_region_name;
       my $desc = "$cs_name $sr_name has 'karyotype_rank' attribute";
       ok($_->has_karyotype, $desc);
-
-      if ($sr_name =~ /^(chrM|chrMT|MT|Mito|mitochondrion_genome)$/) {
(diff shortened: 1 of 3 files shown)


james-monkeyshines:
This is intended to fix the bug described in [ENSPROD-7111](https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7111) and [ENSINT-914](https://www.ebi.ac.uk/panda/jira/browse/ENSINT-914).
If you're wondering why you're a reviewer for this ticket, it's because you are the current relco, and we are introducing a new process to ensure that people don't get surprised by fixed/updated datachecks when handing over dbs... If you have any questions or think these changes are problematic, please let the Production team know, otherwise please approve it (we intend not to merge the PR until everyone has approved, although we might have to be a bit pragmatic if someone is away and the fix is urgent).

ndliberial (approved):

ens-LCampbell (approved):

gnaamati (approved):
Thanks for giving us this update James

ipilizota (approved):
Thank you James :)

This DC does not check Compara databases and I approve it conditionally on positive reviews from the affected teams.

PR #403, opened 2021-09-06
Remove inappropriate compara databases

JAlvarezJarreta:
`SchemaPatchesApplied` DC is not applicable for the rapid release Compara databases (References and Homology annotation at the moment) as they follow a different schema.
After further discussion, some additional DCs have been added for the current Compara RR databases.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
@@ -32,7 +32,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME        => 'CompareSchema',
   DESCRIPTION => 'Compare database schema to definition in SQL file',
-  GROUPS      => ['compara', 'core', 'brc4_core', 'corelike', 'funcgen', 'schema', 'variation'],
+  GROUPS      => ['compara', 'compara_homology_annotation', 'compara_references', 'core', 'brc4_core', 'corelike', 'funcgen', 'schema', 'variation'],
   DB_TYPES    => ['cdna', 'compara', 'core', 'funcgen', 'otherfeatures', 'production', 'rnaseq', 'variation'],
   PER_DB      => 1
 };
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DatabaseCollation.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DatabaseCollation.pm
@@ -30,7 +30,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME        => 'DatabaseCollation',
   DESCRIPTION => 'All tables have the same collation (latin1_swedish_ci)',
-  GROUPS      => ['ancestral', 'brc4_core', 'compara', 'compara_gene_trees', 'compara_genome_alignments', 'compara_master', 'compara_syntenies', 'core', 'corelike', 'funcgen', 'schema', 'variation'],
+  GROUPS      => ['ancestral', 'brc4_core', 'compara', 'compara_gene_trees', 'compara_genome_alignments', 'compara_homology_annotation', 'compara_master', 'compara_references', 'compara_syntenies', 'core', 'corelike', 'funcgen', 'schema', 'variation'],
   DB_TYPES    => ['cdna', 'compara', 'core', 'funcgen', 'otherfeatures', 'rnaseq', 'variation'],
   PER_DB      => 1
 };
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MySQLStorageEngine.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MySQLStorageEngine.pm
@@ -30,7 +30,7 @@ extends 'Bio::EnsEMBL::DataCheck::DbCheck';
 use constant {
   NAME        => 'MySQLStorageEngine',
   DESCRIPTION => 'Database schema matches expected MySQL storage engine',
-  GROUPS      => ['compara', 'core', 'brc4_core', 'corelike', 'funcgen', 'schema', 'variation'],
+  GROUPS      => ['compara', 'compara_homology_annotation', 'core', 'brc4_core', 'corelike', 'funcgen', 'schema', 'variation'],
   DB_TYPES    => ['cdna', 'compara', 'core', 'funcgen', 'otherfeatures', 'rnaseq', 'variation'],
   PER_DB      => 1,
 };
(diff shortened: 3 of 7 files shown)


CristiGuijarro (changes requested):
Requires reindexing index.json file, but otherwise a necessary change for RR

CristiGuijarro (approved):

james-monkeyshines (commented):
I'm not sure that this will work at handover - the 'compara' group will be run then, so will this fail at that point? Or is it only during pipeline execution that this will be a problem?
And if the schema is different, won't [CompareSchema](https://github.com/Ensembl/ensembl-datacheck/blob/release/105/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm) also fail?

Other things may break in less obvious ways, because the datachecks expect 'compara' schemas to all look the same; if the rapid dbs are significantly different, we might need to address that explicitly.

One option would be to create yet another GROUP, say "compara_rapid"; that has the benefit of working in the current infrastructure, but it's a bit annoying, and maybe somewhat bug-prone, to maintain. The handover process would have to be updated to use this group, but we're planning to make the datachecks that are executed dependent on the target server anyway (https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7112).

The alternative is to create a new DB_TYPE, say "compara_rapid" - this perhaps better reflects that we are working with a different beast, and would provide the flexibility to run a subset of datachecks for rapid, and also allow for RR-specific checks. But, this requires a bit more engineering in the background, this won't work out of the box.

CristiGuijarro:
> I'm not sure that this will work at handover - the 'compara' group will be run then, so will this fail at that point? Or is it only during pipeline execution that this will be a problem?
> And if the schema is different, won't [CompareSchema](https://github.com/Ensembl/ensembl-datacheck/blob/release/105/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm) also fail?
> 
> Other things may break in less obvious ways, because the datachecks expect 'compara' schemas to all look the same; if the rapid dbs are significantly different, we might need to address that explicitly.
> 
> One option would be to create yet another GROUP, say "compara_rapid"; that has the benefit of working in the current infrastructure, but it's a bit annoying, and maybe somewhat bug-prone, to maintain. The handover process would have to be updated to use this group, but we're planning to make the datachecks that are executed dependent on the target server anyway (https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7112).
> 
> The alternative is to create a new DB_TYPE, say "compara_rapid" - this perhaps better reflects that we are working with a different beast, and would provide the flexibility to run a subset of datachecks for rapid, and also allow for RR-specific checks. But, this requires a bit more engineering in the background, this won't work out of the box.

As far as I am aware, there will be no "handover" of the per-species compara databases for rapid release. The necessary DCs are incorporated in the pipelines to be run (and these run just the `compara_homology_annotation` group for the blastocyst pipeline and `compara_references` for the reference database update pipeline. The new databases are deposited directly on the rapid release server, with no further checks once they have passed those mentioned.

james-monkeyshines:
> As far as I am aware, there will be no "handover" of the per-species compara databases for rapid release. The necessary DCs are incorporated in the pipelines to be run (and these run just the `compara_homology_annotation` group for the blastocyst pipeline and `compara_references` for the reference database update pipeline. The new databases are deposited directly on the rapid release server, with no further checks once they have passed those mentioned.

Yes, sorry, you're right - I got a bit muddled. However, if we're not running the 'compara' group, then is the 'compara_homology_annotation' one sufficient? There are datachecks that are in the former but not the latter that  (e.g. CompareSchema, MySQLStorageEngine, WhitespaceCritical) which seem equally applicable to standard and RR dbs, and could cause bugs if they fail.

Another thing that has just occurred to me is that the divergence in schema may pose difficulties for us when we increment the Ensembl version for RR - our process is to rename the db with the new version, and then apply schema patches (automated with a script). Can we expect this to work for RR dbs?

CristiGuijarro:
> > As far as I am aware, there will be no "handover" of the per-species compara databases for rapid release. The necessary DCs are incorporated in the pipelines to be run (and these run just the `compara_homology_annotation` group for the blastocyst pipeline and `compara_references` for the reference database update pipeline. The new databases are deposited directly on the rapid release server, with no further checks once they have passed those mentioned.
> 
> Yes, sorry, you're right - I got a bit muddled. However, if we're not running the 'compara' group, then is the 'compara_homology_annotation' one sufficient? There are datachecks that are in the former but not the latter that (e.g. CompareSchema, MySQLStorageEngine, WhitespaceCritical) which seem equally applicable to standard and RR dbs, and could cause bugs if they fail.
> 
> Another thing that has just occurred to me is that the divergence in schema may pose difficulties for us when we increment the Ensembl version for RR - our process is to rename the db with the new version, and then apply schema patches (automated with a script). Can we expect this to work for RR dbs?

We could include `CompareSchema` and `WhiteSpaceCritical` in `compara_homology_annotation`  - I see no issue there, I must have missed those when creating the new groups initially. Should @JAlvarezJarreta include those in this PR?

JAlvarezJarreta:
I agree with you @james-monkeyshines the new DB_TYPE may actually fit best with how we are proceeding with our rapid release development (for now). But I think due to the need of homology annotation to be out asap (and the additional development required), this can be discussed in a meeting if you want (so the next compara RR pipelines are adapted to this approach). I'm happy to add those DCs you both seem deemed for these new compara databases in thir PR.

JAlvarezJarreta:
@CristiGuijarro, I will add those two DCs to `compara_references` since they seem useful as well, what do you think? I have added `MySQLStorageEngine` only to `compara_homology_annotation` as I feel this DC should be only applied to release databases.

james-monkeyshines:
Hi @JAlvarezJarreta, @CristiGuijarro, thanks for your comments/feedback. The three datachecks I highlighted were intended to be examples, rather than an exhaustive list - I've had another look at the generic datachecks (I'm assuming the more compara-y ones are fine), and I think DatabaseCollation and SchemaVersion would also be useful to have.

CristiGuijarro:
> Hi @JAlvarezJarreta, @CristiGuijarro, thanks for your comments/feedback. The three datachecks I highlighted were intended to be examples, rather than an exhaustive list - I've had another look at the generic datachecks (I'm assuming the more compara-y ones are fine), and I think DatabaseCollation and SchemaVersion would also be useful to have.

Can definitely include `DatabaseCollation`, and we could include `SchemaVersion` for now too -  but this may require change in the future (for example with the decoupled schema, we have not yet come up with a schema versioning system for the RR compara db - and I believe the db name will still need to fit with the main release versioning).

CristiGuijarro (approved):
:+1:  I think that should do it for now :crossed_fingers:

james-monkeyshines (approved):

james-monkeyshines:
Just spotted that this is against the master branch - for RR we run against the latest release branch for datachecks, ie release/105.
Are these changed needed right away? If so, I'll merge into the release branch (which will be merged back into master in the next week or two in any case).

JAlvarezJarreta:
Oh apologies! Yes, @luca-drf  is running the homology annotation pipeline atm so I guess he needs these changes asap. Thank you again @james-monkeyshines!

JAlvarezJarreta:
I have created a new PR with the right branches (see PR#405).

PR #404, opened 2021-09-08
Enforce INSDC accessions as seq_region names or synonyms.

james-monkeyshines:
For new species, and most old ones, we have INSDC accessions for all of our sequences - this is really useful for users. Due to exceptional cases on the main Ensembl site, however, we don't enforce INSDC accessions, and the relevant datacheck was advisory.

For Rapid Release, however, this data should always be available, and we are introducing processes which depend on this (ENA-format GFF3 dumps, and the new metadata schema).

This PR makes the datacheck critical for RR, remaining advisory for the main site.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionNamesINSDCAdvisory.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/SeqRegionNamesINSDCAdvisory.pm
@@ -16,7 +16,7 @@ limitations under the License.
 
 =cut
 
-package Bio::EnsEMBL::DataCheck::Checks::SeqRegionNamesINSDC;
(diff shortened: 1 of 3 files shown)


james-monkeyshines:
For this datacheck to be executed at handover, this development is required: https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7112

marcoooo (approved):

vsitnik (approved):
Seems to be ok for the recently loaded metazoan cores.
Older ones from the main site won't pass this check.

Approving.

james-monkeyshines:
Should have mentioned before, the datacheck passes for current RR species, with the exception of 4 species about which the relevant teams have already been notified.

gnaamati (approved):
Thanks James, we will be fixing the relevant cores in RR

ndliberial (approved):

PR #409, opened 2021-09-16
Add common name check to core_sync group.

james-monkeyshines:
Checking common names across a range of databases is awkward to do at handover, but things break if there are discrepancies, so we certainly want it as a critical datacheck.

The datacheck logically fits in the existing 'core_sync' group, which is convenient because we already run that as a post-handover pipeline, so don't need any additional process or config to incorporate the check.


james-monkeyshines:
The failure here is due to travis config, which should be fixed in #410.

marcoooo (approved):

PR #411, opened 2021-09-17
EP-7007: Fix failures due to LRG insertion contigs

luca-drf:
Fix ENSPROD-7007 ignoring changes in LRG regions


marcoooo:
@luca-drf were you able to test this update against a failing human db?

luca-drf:
> @luca-drf were you able to test this update against a failing human db?

@marcoooo Yes, I've tested it on human 104 vs human 105

luca-drf:
> Another question is whether we would like to see this integrated to 106 branch, therefore changing target branch.

@marcoooo according to Havana, there won't be any LRG insertions for 106 so we don't need that on 106 release branch.

marcoooo (approved):

PR #415, opened 2021-09-22
Dispalynameformat skip for main site

vinay-ebi:
Mis-format display name will afftect the new ftp file dumping pipeline used in rapid relase. There is a Datacheck DisplayNameFormat  to Find this error. We can add this dc to core group to find out error during initial stage of handover but it will enforce to apply for main site as well which is not mandatory for time being. The new skip_test  method will allow to apply this dc for rapid relase alone.

Moreinfo in ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSPROD-7112


james-monkeyshines (commented):
This change isn't necessary; we already have the functionality we need.
By only having the 'rapid_release' group defined, this datacheck is excluded from the main site handover. It's only by re-introducing the 'core' group that this becomes a problem. We do not want to use skip_tests in this way, with hard-coded references to our server names. So it's best just to leave it as is.

PR #426, opened 2021-11-03
Fix Schema DCs to fit edgecases

CristiGuijarro:
## Description
There were failing schema related DCs for the compara RR databases. Whilst fixing this on compara side it was noticed that the DCs themselves needed slight tweaks too.
* `CompareSchema` DC was missing a particular syntax for `UNIQUE KEY` and `CONSTRAINTS` so this line has been added.
* `MySQLStorageEngine` was hardcoded to MyISAM engine check, which is fine for every database except the compara RR databases, which require InnoDB engine, so a ternary has been chucked in to fit.

## Jira Ticket:
[ENSCOMPARASW-4959](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-4959)

## Related PR
https://github.com/Ensembl/ensembl-compara/pull/363 <- for interest/further understanding

## Testing
- Tested on RR db on `InnoDB` schema
- Tested on e104 and e105 main release compara

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
@@ -181,6 +181,7 @@ sub normalise_table_def {
 
   # Remove keys from table definition since they have been extracted.
   $table =~ s/^((?:PRIMARY |UNIQUE |FOREIGN )*KEY\s.*)\n//gm;
+  $table =~ s/^((?:CONSTRAINT |UNIQUE |KEY ).*)\n//gm;
 
   my ($table_name) = $table =~ /CREATE TABLE (\S+)/m;
 
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MySQLStorageEngine.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MySQLStorageEngine.pm
@@ -38,7 +38,7 @@ use constant {
 sub tests {
   my ($self) = @_;
   my $database_name = $self->dba->dbc->dbname;
-  my $engine = 'MyISAM';
+  my $engine = ($database_name =~ /_compara_/ and $database_name !~ /ensembl/) ? 'InnoDB' : 'MyISAM';
   my $diag = "Non-$engine table";
   my $desc = "All tables are using MySQL $engine storage engine";
   my $sql = qq/ SELECT TABLE_NAME FROM 


lib/Bio/EnsEMBL/DataCheck/Checks/MySQLStorageEngine.pm
```diff
@@ -38,7 +38,7 @@ use constant {
 sub tests {
   my ($self) = @_;
   my $database_name = $self->dba->dbc->dbname;
-  my $engine = 'MyISAM';
+  my $engine = ($database_name =~ /_compara_/ and $database_name !~ /ensembl/) ? 'InnoDB' : 'MyISAM';
```
marcoooo, line 41: Not sure about this one, the MyISAM storage engine is compulsory for subsequent processing, why should we accept `innoDB` for compara particularly?

CristiGuijarro: This is specific to the rapid release compara databases and not "main" release compara databases. When you say subsequent processing - what exactly do you mean, what subsequent processing is performed on the per species RR compara dbs? The decision for these dbs has been to be `InnoDB` for a while - and by default for RR they were on our side, but in the handover process, this seems to have been lost. In testing, there did not appear to be any issues with the use of `InnoDB` with web - but perhaps they did not notice?

CristiGuijarro: A further note on the use of `InnoDB` in the per species compara dbs instead of `MyISAM` is that the schema takes advantage of the `INDEX`, `FOREIGN KEYS` and `CASCADE`, in some ways that the "main" release compara db cannot - with `MyISAM`

CristiGuijarro: Sorry :sweat_smile: , that was not intended to teach you the benefits of `InnoDB` over `MyISAM` - I was trying to highlight that the specific features of `InnoDB` in the RR compara db depends on these in ways that the "main" release db at present does not. Here I filter on the name to match `compara` because this is a key word in the compara dbs, but exclude the use of `ensembl` - because this is in the nomenclature of the main release compara dbs - so "main" release dbs *should* not slip through.

marcoooo: ok got it.

lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
```diff
@@ -181,6 +181,7 @@ sub normalise_table_def {
 
   # Remove keys from table definition since they have been extracted.
   $table =~ s/^((?:PRIMARY |UNIQUE |FOREIGN )*KEY\s.*)\n//gm;
+  $table =~ s/^((?:CONSTRAINT |UNIQUE |KEY ).*)\n//gm;
```
marcoooo, line 184: May be  the `foreign` should still be kept in case we have a compara MyISAM right? I mean this is somehow the same consequence about the ENGINE tests, am I mistaking?

CristiGuijarro: These keys are collected and stored earlier in the script to a hash of keys, they are only removed here for the remainder of the comparison in the `is_deeply` `test`. In this case the above regex was simply missing the removal of the `KEY`s which followed a differing syntax either in the `table.sql` schema or the schema collected directly from the - for example, those that do not end in `KEY` and can start with `CONSTRAINT` - I did try to fix the above regex, but I was worried about breaking it for something I could not test and it seemed the safer option just to include another substitution regex. Is that okay?

marcoooo (approved):

PR #433, opened 2022-01-07
Merge latest updates from 105 back to main before 107 branching. 

marcoooo:


marcoooo:
@thibauthourlier  @JAlvarezJarreta I added you to make sure I won't put back a regression in main before branching to 107.

PR #435, opened 2022-01-10
Allow for _ to be part of the display name format.

ens-carlos:
Example: Xenopus laevis (J_2021) - GCA_017654675.1


marcoooo:
Hi @ens-carlos Since I guess this is for Rapid Release, I guess this should be applied to 105 release branch as well, right?

ens-carlos:
Yes, that's right.

marcoooo (approved):

PR #436, opened 2022-01-10
Allow for _ to be part of the display name format.

ens-carlos:
Example: Xenopus laevis (J_2021) - GCA_017654675.1


marcoooo:
Merged and pulled on RR branches. @ens-carlos you may resubmit your dbs then.

PR #454, opened 2022-01-26
updated method to add new genebuild metakeys

ens-ftricomi:
Updated regex for method in VersionedGenes.pm in order to include the new genebuild pipelines (braker, anno, standard).
Marc, could you please double check that this will not affect any Production code?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,19 +18,26 @@ addons:
     - graphviz
 
 before_install:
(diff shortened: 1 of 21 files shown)


marcoooo:
@ens-ftricomi  You have much more than just the expected update in there. Could you resubmit with only the patch on the DC regexp?

PR #457, opened 2022-01-26
EP-7110: Refactor HGNCMultipleGenes to allow comparison with threshold

luca-drf:
Refactor `HGNCMultipleGenes` into `CompareHGNCMultipleGenes`.

The new DC will fail only if the increase in HGNC symbols assigned to multiple genes compared to the previous release is above 5%.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CompareHGNCMultipleGenes.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CompareHGNCMultipleGenes.pm
@@ -16,7 +16,7 @@ limitations under the License.
 
 =cut
 
-package Bio::EnsEMBL::DataCheck::Checks::HGNCMultipleGenes;
+package Bio::EnsEMBL::DataCheck::Checks::CompareHGNCMultipleGenes;
 
 use warnings;
 use strict;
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/CompareHGNCMultipleGenes.pm
```diff
@@ -16,45 +16,65 @@ limitations under the License.
...
+  my $cur_multi_hgnc_percentage = $self->multiple_hgnc_percentage($cur_dba);
+  my $old_multi_hgnc_percentage = 0;
+
+  if (defined $old_dba) {
+    $old_multi_hgnc_percentage = $self->multiple_hgnc_percentage($old_dba);
+  }
+
+  my $multi_hgnc_diff = $cur_multi_hgnc_percentage - $old_multi_hgnc_percentage;
+
+  my $desc = "HGNC symbols assigned to multiple genes have not increased more than 5%";
```
marcoooo, line 53: Do we expect HGNC symbol to increase by more than 5% every release?

luca-drf: Genebuild suggested to fail the test when the proportion of HGNC symbols assigned to multiple genes increases more than 5% from the previous release.

marcoooo: Was just wondering if a drop had the same potential issue.

marcoooo (approved):

lib/Bio/EnsEMBL/DataCheck/Checks/CompareHGNCMultipleGenes.pm
luca-drf: I did wonder the same thing, but I'm guessing it's not a problem to have several less symbols assigned to multiple genes. Also, this a very specific test and we have several other checks in place for HGNC and xrefs, so it's probably okay. I think we can ping Genebuild again and tune the threshold or add another check if we find that this one is too permissive.

PR #461, opened 2022-02-17
Hotfix/genes biotypes

thibauthourlier:
* Removed useless disconnect as a connection will be needed for every get_all_Transcripts call
* use slices to fetch genes instead of fetching all the genes
* Check key existence when looking for coding transcripts in a pseudogene gene as for me it is faster and similar to looping through all the keys of the hash and to do a string comparison


thibauthourlier:
I'll create it properly against 107

PR #464, opened 2022-03-14
Allow for apostrophe's in display name, e.g. "Geoffroy's cat"

leannehaggerty:
DC was failing for display name:
Leopardus geoffroyi (Geoffroy's cat) - GCA_018350155.1

Added apostrophe to regex

Have requested Anne review to make sure this won't cause problems for web


ens-ap5 (approved):
I don't think this will cause problems for us, since we typically double-quote strings with values in them, but worth checking on the RR test site when this change goes in, just to be 100% sure.

marcoooo (approved):

PR #465, opened 2022-03-25
new datacheck added to check meta key strain group for rapid release …

vinay-ebi:
Disallow species.strain_group meta key in core dbs for rapid release


marcoooo (approved):
All good, can you open the exact same one against 106 please? Since this is the next target release for RR.

PR #468, opened 2022-04-01
added exception for stable id prefix

ens-ftricomi:
The  otherfeatures db usually has RefSeq import with its own stable id prefix or cDNA without the stable id, the stable id prefix meta key if present follows the one the core db. In the new genebuild pipeline the otherfeature db  hasn't the same stable id prefix that is present in the core db because the annotation is provided by Braker. So we have two stable is prefix.
Will it cause any issue for Production/Web? 
Can we add an exception for the stable id prefix in order to pass the  DCs?


leannehaggerty (commented):
This is all good from our side, so unless web use this key in the otherfeatures db, I think the change in the DC will be fine

ens-ftricomi:
Not needed anymore, I had a chat with Thibaut about it and I will leave the ENS prefix even in the otherfeatures db

PR #471, opened 2022-04-21
New DC for checking that value for strain.type meta key is from approved list

leannehaggerty:
This is my first DC from scratch so I apologise for any mistakes, but the test will be on Genebuild when we handover databases for RR34, if any issues pop up we will fix them immediately.

**Description**
We found that the value for strain.type meta key had been set incorrectly as "reference" for a db on RR, this DC should check that the value for strain.type can only be from the approved list (we can update this list as needed).


ens-ap5 (approved):
The default list looks good to me, but I know nothing about data check syntax :)

marcoooo (changes requested):
The DC doesn't seem to be registered in index.json file.  therefore this DC would never be run.

leannehaggerty:
Thanks @marcoooo, I think it's ok now

marcoooo (approved):

PR #486, opened 2022-07-04
Added division path to check data_files availability.

marcoooo:
With data_file initial directory updates, the DC need updating to include division name in path.


pblins (approved):
DC was failing because division name is part of the funcgen data_files full path. 
Everything is working now.

PR #490, opened 2022-08-18
Bug/fix old db uri

vinay-ebi:
details are in ticket : ENSPROD-7996


marcoooo (changes requested):
Seems all good, but you might need to change automated tests.

marcoooo (approved):

PR #491, opened 2022-08-18
Changes to support EVA datachecks

olaaustine:
[ENSVAR-5000](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5000)
This datacheck is intended to make sure VariationClasses check is done based on the source of the variants.
The idea is if EVA is present in the source table then import was done using EVA so only EVA checks, if not present dbSNP check is done.


dglemos (approved):
The update looks good.

PR #495, opened 2022-08-26
test case for multi old_db_uri

vinay-ebi:


marcoooo:
@vinay-ebi Build is passing, should we consider moving this to "ready for review" then?

vinay-ebi:
> @vinay-ebi Build is passing, should we consider moving this to "ready for review" 
Need to include one more test case to cover the array of old server URI

vinay-ebi:
Open against release/110

PR #497, opened 2022-09-13
For projection builds: Allow up to 100 gene with without a transcript that has a matching biotype

leannehaggerty:
To allow for human projections to pass DCs we need to allow a small number of genes to have transcripts that have different biotypes

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/GeneBiotypes.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/GeneBiotypes.pm
@@ -136,9 +136,18 @@ sub biotype_groups {
     }
   }
 
-  my $desc_1 = 'genes have at least one transcript with a matching biotype group';
-  is(scalar(@group_mismatch), 0, $desc_1);
-
+  my $mca = $self->dba->get_adaptor('MetaContainer');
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/GeneBiotypes.pm
```diff
@@ -136,9 +136,18 @@ sub biotype_groups {
     }
   }
 
-  my $desc_1 = 'genes have at least one transcript with a matching biotype group';
-  is(scalar(@group_mismatch), 0, $desc_1);
-
+  my $mca = $self->dba->get_adaptor('MetaContainer');
+  if($mca->single_value_by_key('genebuild.method') eq 'projection_build'){
+      if (scalar(@group_mismatch) > 100){
+	  my $desc_proj = 'genes have at least one transcript with a matching biotype group';
```
marcoooo, line 142: I would change the message to something more meaningful there:
```suggestion
	  my $desc_proj = 'projected genes have at most 100 transcripts with a matching biotype group';
```
may be?

leannehaggerty: updated to 'found no more that 100 genes with mismatched transcript biotype groups'

marcoooo (approved):

PR #508, opened 2022-10-12
Adding annotation source tag to the display name

vsitnik:
Allowing appending "[tags]" to the display name


marcoooo:
@vsitnik I need the same against 108 and 109 please

marcoooo (approved):

PR #514, opened 2022-10-28
Hotfix/check mt karyotype

thibauthourlier:
Check if the mitochondria has the karyotype_rank attribute when chromosomes are presents, i.e other regions have a karyotype_rank attribute.
Check if the mitochondria does not have the karyotype_rank attribute when no chromosomes are present
This is linked to ENSINT-1417


thibauthourlier:
Wrong branch to do the PR against...

PR #521, opened 2023-01-09
Update MitochondriaAnnotated.pm

vsitnik:
Some times we have genomes where only MT is a chromosome.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MitochondriaAnnotated.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MitochondriaAnnotated.pm
@@ -79,7 +79,7 @@ sub tests {
       # It shouldn't have the attribute in any other case
       my $chromosomes = $sa->fetch_all_karyotype;
       my $karyotype_rank = $slice->karyotype_rank;
-      if (@$chromosomes > 1) {
+      if (@$chromosomes > 0) {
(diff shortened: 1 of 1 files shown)


thibauthourlier:
The reason for the DC to fail if the MT is the only chromosome in the genome is that on the website, it would appear in the Whole genome/karyotype view. It is not great if users can go on the whole genome view and only see a MT. I could be wrong in how the current and "new" website will display the karyotype.

I don't see a problem in having a karyotype_rank value for only one chromosome but it does feel strange.

I think someone from Web, Infrastructure and Outreach should also be contacted to make sure that what I say in not just my belief.

vsitnik:
We have this situation for ages and it was ok for us (metazoa), web, infrastructure and outreach.
This blocks our e110 handover. We need to remove this constraint. And then we can check later with everyone, I assume.

vsitnik:
Another thought.
We have other species on the main  rapid site. And we're not sure that removing MT KR from the cores we already have in prod (main, rapid) won't affect users or our processes (production and compara pipelines)

thibauthourlier:
We never had any problems with Production or Compara when our scaffold based assemblies had a MT without the karyotype rank. Maybe Productioin or Compara will disagree 😂

thibauthourlier (approved):
I really do think only chromosome level assemblies should have the karyotype_rank attributes, but I won't stop people from working on my last month

marcoooo (approved):

PR #523, opened 2023-01-25
Generalise CheckComparaStableIDs prefix subtest

twalsh-ebi:
Compara datacheck `CheckComparaStableIDs` has a gene-tree stable-ID prefix consistency check which compares the stable ID prefixes of a given division against a known prefix.

This fails if run on any gene-tree prefix not included in the hash used by the datacheck to store known stable ID prefixes (e.g. WormBase gene trees with stable ID prefix `WBGT`). To avoid having to update the datacheck for every gene-tree prefix that needs to be supported, this PR generalises the stable ID prefix subtest into two subtests: one to check that all relevant gene tree stable IDs follow the standard format (described in the [Compara documentation on gene tree stable IDs](https://www.ensembl.org/info/genome/compara/stable_ids.html) and implemented in the [Compara::StableId::Generator method generate_new_name](https://github.com/Ensembl/ensembl-compara/blob/1eaea8d0de9491b3e7d37b1d423cb6c7552bed90/modules/Bio/EnsEMBL/Compara/StableId/Generator.pm#L89-L95)), and a second subtest to check that the prefix is consistent across all gene trees having a standard-format gene tree stable ID.

The new version of the datacheck uses a Perl regex to extract the stable ID prefixes. The datacheck was tested on Fungi Compara 107 and 108 as well as a WormBase gene-trees pipeline to confirm that this regex solved the problem without introducing a second problem.

### Testing

Relevant subtest output for Fungi Compara 107 (with some default gene trees having standard stable IDs):
```
not ok 3 - All gene-tree stable_ids follow the standard format
#   Failed test 'All gene-tree stable_ids follow the standard format'
#   at xxxxxx/ensembl-datacheck/lib/Bio/EnsEMBL/DataCheck/Checks/CheckComparaStableIDs.pm line 79.
#          got: '50921'
#     expected: '0'
ok 4 - There is a single consistent prefix for all standard-format gene tree stable_ids
```

Relevant subtest output for Fungi Compara 108 (with all default gene trees having standard stable IDs):
```
ok 3 - All gene-tree stable_ids follow the standard format
ok 4 - There is a single consistent prefix for all standard-format gene tree stable_ids
```

Relevant subtest output for a WormBase gene-tree pipeline:
```
ok 3 - All gene-tree stable_ids follow the standard format
ok 4 - There is a single consistent prefix for all standard-format gene tree stable_ids
```

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CheckComparaStableIDs.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CheckComparaStableIDs.pm
@@ -56,28 +56,35 @@ sub tests {
   /;
   is_rows_zero($self->dba, $sql_2, $desc_2);
 
-  my %prefixes = (
-    "vertebrates" => "ENSGT",
-    "plants"      => "EPIGT",
-    "pan"         => "EGGT0",
-    "metazoa"     => "EMGT0",
-    "protists"    => "EPrGT",
-    "fungi"       => "EFGT0"
-  );
-
-  my $division = $self->dba->get_division();
-  my $prefix = $prefixes{$division};
-
-  my $desc_3 = "There is a single consistent stable_id prefix for all gene trees";
+  my $desc_3 = "All gene-tree stable_ids follow the standard format";
   my $sql_3 = qq/
-    SELECT * FROM gene_tree_root
+    SELECT stable_id FROM gene_tree_root
       WHERE member_type = 'protein'
         AND tree_type = 'tree'
         AND clusterset_id =  "default"
-        AND LEFT(stable_id, 5) != "$prefix"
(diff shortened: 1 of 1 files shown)


marcoooo:
Hi @twalsh-ebi , thank you for this, although are Fungi part of Wormbase? I would have preferred a test on target initial problematic solved.

twalsh-ebi:
> Hi @twalsh-ebi , thank you for this, although are Fungi part of Wormbase? I would have preferred a test on target initial problematic solved.

That's a fair point @marcoooo. I've run a datacheck on a WormBase pipeline to confirm that it works as expected on the motivating use case, and I've updated the PR description with example output of the relevant subtests.

marcoooo (approved):

PR #528, opened 2023-02-16
Run Alphafold checks only once for multispecies DB

EbiArnie:
Alphafold inserts features without regard for the species.
The datacheck would check the same thing for every species in the collection.
Put a PER_DB = 1 to make it run only once for collections.


marcoooo (commented):
I wonder if this test shouldn't be updated further, because in case you didn't assign any alphafold, it would fail. but is this something that can actually happen?

marcoooo (approved):

EbiArnie:
> I wonder if this test shouldn't be updated further, because in case you didn't assign any alphafold, it would fail. but is this something that can actually happen?

Could be but unlikely. So:
* A species usually has transcripts that are protein coding
* It will then usually have a Uniprot accession
* For this we will usually have an Alphafold accession

All of these are very likely, but not "must".

I'd keep it like it is, because the datachecks should catch "probably wrong" situations.

PR #529, opened 2023-02-17
New DC to check for Gene Symbol Classifier in Ensembl cores

leannehaggerty:
This will check that the GSC has been run on cores that have ensembl annotations (it will skip anything that is not 'anno' or 'full_genebuild')

I am still in the process of handing over the cores that previously failed in the gene symbol classifier pipeline, once handed over there shouldn't be anything on sta-5 that fails the DC


lib/Bio/EnsEMBL/DataCheck/Checks/GeneSymbols.pm
```diff
@@ -0,0 +1,56 @@
+=head1 LICENSE
+Copyright [2018-2022] EMBL-European Bioinformatics Institute
```
marcoooo, line 2: License should be 2023.

marcoooo (changes requested):

lib/Bio/EnsEMBL/DataCheck/Checks/GeneSymbols.pm
```diff
@@ -0,0 +1,56 @@
...
+     if($method ne 'anno' && $method ne 'full_genebuild') {
+         skip "Gene Symbols not mandatory for non-Genebuild annotations", 1;
+     }
+
+     my $desc = 'Gene Symbol Transformer has been run';
+     my $sql  = qq/
+      SELECT COUNT(*) FROM
+        analysis
+      WHERE
+        logic_name LIKE 'gene_symbol_classifier'
```
ens-LCampbell, line 50: Is there any other variation of logic name related to gene_symbolc_classifier expected ?  'gene_symbol_classifier%' ? 

If not, would be better to be explicit:
logic_name = 'gene_symbol_classifier'.

lib/Bio/EnsEMBL/DataCheck/Checks/GeneSymbols.pm
leannehaggerty: Yep, sorry fixed that

ens-LCampbell (approved):

lib/Bio/EnsEMBL/DataCheck/Checks/GeneSymbols.pm
leannehaggerty: We're likely going to add more specific classifiers, i.e. the current is trained on verts but we're working on a non-verts model now. I'll need to update this DC in the future either way

ens-LCampbell (approved):
Code checks out for me, happy to approve merge

marcoooo (approved):

PR #546, opened 2023-03-30
Change VariationSetStructure to be advisory

dglemos:
Details: ENSVAR-4313


marcoooo (changes requested):
you might want to add `DATACHECK_TYPE => 'advisory',` in the DC itself, to fix the build.

marcoooo (approved):

PR #557, opened 2023-08-22
Added datacheck for timely release of eHive semaphors

ens-sb:
# Description of the problem
Early semaphore release issues affected several pipelines in e111. Such issues have been reported before, in e109 Vertebrates FTP dumps, [e105 Mammals EPOwithExt](https://genomes-ebi.slack.com/archives/D0255PFSNKW/p1625849995012100), [e103 default ncRNAtrees](https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/Intentions+for+release+103), and in [e96 Plants](https://www.ebi.ac.uk/seqdb/confluence/display/EnsCom/2019-02-27).

But the impact of this issue in e111 — one pipeline completely rerun, a second pipeline with out-of-sync homology data — calls for measures to catch such issues as soon as possible.

# Scope of the pull request

The aim of this PR is to create a datacheck (**TimelySemaphoreRelease**) which queries the semaphore and job tables to check that every dependent semaphore job was started only after all of its fan jobs had completed. The datacheck should fail if this is not the case for any semaphored job.
Relevant compara ticket: [ENSCOMPARASW-6845](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-6845)

*WARNING: the position of the CompareVariationRows index entry was changed by this PR.*

# Testing

The datacheck was manually run on two databases where failure is expected and one where pass is expected.

PR #558, opened 2023-08-22
Add datacheck MemberXrefAccessTime

twalsh-ebi:
## Description of the problem

In Ensembl 109 the access time of Member Xref data from the Metazoa Compara database was prohibitively slow, causing gene-tree displays to fail with an Ajax error ([ENSWEB-6789](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6789)).

It has been proposed to develop a datacheck ([ENSCOMPARASW-6449](https://www.ebi.ac.uk/panda/jira/browse/ENSCOMPARASW-6449)) to catch and fix this kind of issue at an earlier stage in the release cycle.

## Scope of the pull request

This PR adds a datacheck (`MemberXrefAccessTime`) which tests that access of Member Xref data via the Compara Perl API is not prohibitively slow.

This datacheck is added to the `compara_annot_highlight` group so that it can be run as soon as the `member_xref` and `external_db` tables of the relevant Compara database have been populated.

Note that in the datacheck config, the `TABLES` array is left empty in case an issue arises with Member Xref access of data that has not changed.

_WARNING: like #557, this PR moves the CompareVariationRows datacheck index entry so that it is placed alphabetically._

## Testing

The datacheck was run on 2 example Metazoa Compara databases: one unmodified database and one database with the primary key of the `member_xref` table dropped.

As expected, the datacheck passed on the former database and failed on the latter.

---

The datacheck was also run on `ensembl_compara_fungi_57_110`, and passed in 21 minutes and 27.83 seconds.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/MemberXrefAccessTime.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/MemberXrefAccessTime.pm
@@ -66,8 +66,7 @@ sub tests {
     LIMIT 1;
   /;
 
-  my $external_db_names = $helper->execute_simple(-SQL => $external_db_query);
-  my $external_db_name = $external_db_names->[0];  # only one external db is needed
+  my $external_db_name = $helper->execute_single_result(-SQL => $external_db_query);
 
 
   my $avg_tree_size_query = q/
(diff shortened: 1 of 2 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/MemberXrefAccessTime.pm
```diff
@@ -0,0 +1,138 @@
...
+  my $external_db_query = q/
+    SELECT
+      db_name
+    FROM
+      external_db
+    LIMIT 1;
+  /;
+
+  my $external_db_names = $helper->execute_simple(-SQL => $external_db_query);
+  my $external_db_name = $external_db_names->[0];  # only one external db is needed
```
ens-sb, line 70:
```suggestion
  my $external_db_name = $helper->execute_single_result(-SQL => $external_db_query);
```

ens-sb (approved):

PR #570, opened 2024-01-30
Update TranscriptDisplayXrefSuffix.pm

marcoooo:
This test keeps failing for some Plants Xref - but this is because the new Xref pipeline is in used. 
So deactivated this test for plants division (seen with @gnaamati) 
Thanks!


dpopleton (approved):

gnaamati (commented):
I'm ok with this

PR #573, opened 2024-02-08
Add support to GCF to ease moving core dbs between beta/rapid and main

JAlvarezJarreta:
We tend to release in both beta/rapid and then main site some of our organisms, so users can access core information quickly and then add other relevant features such as CACTUS alignments and protein trees. To ease us moving core databases from one system to the other, it would be helpful to add support for GCF values in `assembly.accession` meta key.

_Scope:_ e113, thus the PR against `main`


JAlvarezJarreta:
Support of GCF in main sites rejected.

PR #589, opened 2024-06-28
Deprecate RegulatoryFeatureIsActive.pm DC

pblins:
Deprecate RegulatoryFeatureIsActive.pm (checks if all reg. features are active).


vinay-ebi:
open against the release/113

PR #623, opened 2025-02-26
Update PolyploidAttribs datacheck

JAlvarezJarreta:
- Skip test only when no indication of ploidy is found
- Added check of ploidy meta key value
- Improved overall performance of DC

Tested in several polyploid and non-polyploid core databases in st3, some of which will need a patch to pass this updated DC (as expected with the improved checks).


vsitnik (approved):
Thank you!

dpopleton (approved):

PR #624, opened 2025-02-26
Bugfix: exclude "synthetic" Un chromosomes in PolyploidAttribs

JAlvarezJarreta:
_T. aestivum_ and _T. turgidum_ have a "synthetic" chromosome named `Un` introduced to gather all the scaffolds not assigned to actual chromosomes so they belong to the same genome component, `U`. This produces the actual count to be +2 over the actual ploidy, causing this DC fail for the two aforementioned species.

The correction applied is to skip based on chromosome name rather than genome component, as it is possible for an actual genome component `U` to exist, but it would be very unusual to find a chromosome named `Un`.

_Note:_ Thank you @twalsh-ebi for spotting this issue!


vinay-ebi:
Hi @JAlvarezJarreta is this DC still valid for 116 ?

JAlvarezJarreta:
@vinay-ebi Yes please!

vinay-ebi (approved):

PR #625, opened 2025-03-05
Update DbDataChecks_conf.pm

vinay-ebi:
Remove explicit declaration of hive version


nwillhoft (commented):
Out of curiosity, are we good to remove the explicit version because we control that in the modenv?

Yes, in Modenv hive version updated to 2.8.0, This new system cannot validate the minor version 

perl -E 'use Bio::EnsEMBL::Hive::Version 2.8;'
Bio::EnsEMBL::Hive::Version version 2.8 required--this is only version 2.8.0 at -e line 1.
2.8 parses as 2.800, 2.8.0 would parse as 2.8.0. "v2.8" parses as 2.8
 
No proper fix as of now this works

nwillhoft (approved):
Changes look good to remove explicit hive version.

PR #631, opened 2025-04-24
Revert "fix keyword capitalisation in column definitions for CompareSchema DC"

dpopleton:
Reverts Ensembl/ensembl-datacheck#630

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
@@ -110,7 +110,7 @@ sub normalise_table_def {
   $table =~ s/ IF NOT EXISTS//gm;
 
   # Normalise case: everything after column name is upper-cased.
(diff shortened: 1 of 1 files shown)


jgtate (approved):

lib/Bio/EnsEMBL/DataCheck/Checks/CompareSchema.pm
```diff
@@ -110,7 +110,7 @@ sub normalise_table_def {
   $table =~ s/ IF NOT EXISTS//gm;
 
   # Normalise case: everything after column name is upper-cased.
-  $table =~ s/^([a-zA-Z]\w+\s)(.+)/$1\U$2/gm;
```
vinay-ebi, line 113: If we want the column to be lowered in cases, we need to remove the \U for matched group 2

vsitnik: \U is for keywords following the column name
 I'll add a similar fix, but that one will be capitalizing before  ' are torn off.
Something like if m/^\s*'.../ (By ' I mean backticks)

PR #633, opened 2025-04-28
Update CompareSchema for column name one

vinay-ebi:
Existing Regex expect the column name to be greater than one which fails for compara homology table which contains column name `n` and `s` 

$table =~ s/^([a-z]\w+\s)(.+)/$1\U$2/gm; changed to $table =~ s/^([a-z]\w*\s)(.+)/$1\U$2/gm;


dpopleton (approved):
I will trust you.

PR #644, opened 2025-08-05
Exclude Alpha fold entries from the protein feature datacheck

vinay-ebi:
Protein features do not extend beyond the bounds of the translated sequence. This test is an exception for alpha fold entries, which we store in the same table.

 This  issue was resolved for the main site in PR:  https://github.com/Ensembl/ensembl-datacheck/pull/593


nwillhoft (approved):
Changes look good. As mentioned in PR, this code is used and tested on the current site. This will get used in beta.

PR #654, opened 2025-10-24
add helixer to allowed annotation source list

JackCurragh:
Added helixer to the list of allowed annotation sources


leannehaggerty (approved):
LGTM
Thanks Jack

PR #656, opened 2025-11-20
check Beta meta key rather than main site beta key for sample gene 

JackCurragh:
This is a simple DC update for Beta where we just use a new meta_key/value pair for the sample gene


leannehaggerty (approved):
Thanks Jack. LGTM
Just noting that it would be good to extend this check (or create a new one) to make sure the genebuild.sample_location has the correct information - should be a separate PR either way.

JAlvarezJarreta (approved):
Looks good, and the meta key change makes sense.

One thing we were discussing in Plantazoa in the past is the way this sample gene is selected: we were wondering if selecting one from our side at loading time, but then having it reviewed/changed once Compara, Variation and Regulation data is present would be a more meaningful and valuable path to follow for our users. The current set up resulted in some sample genes without Compara data associated with them in the main sites, which is not great IMO.

PR #657, opened 2025-11-21
Fix/check beta sample location meta value

JackCurragh:
Added set of DCs for `genebuild.sample_location`

- Check value exists just once
- Then formatting is <string>:<start number>-<end number> which I figured is safe enough.
- Then that start number < end number
- Then one query that checks if the seq region specified is valid (a join on Name and region_start > 1 and region end < sr.length then check for NULL)
- Then lastly I check if there is a gene in the sample_location

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
@@ -1,6 +1,6 @@
 =head1 LICENSE
 
-Copyright [2018-2024] EMBL-European Bioinformatics Institute
+Copyright [2018-2025] EMBL-European Bioinformatics Institute
 
 Licensed under the Apache License, Version 2.0 (the 'License');
 you may not use this file except in compliance with the License.
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
@@ -1,0 +1,117 @@
+=head1 LICENSE
+
+Copyright [2018-2024] EMBL-European Bioinformatics Institute
+
+Licensed under the Apache License, Version 2.0 (the 'License');
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an 'AS IS' BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package Bio::EnsEMBL::DataCheck::Checks::DisplayableSampleLocation;
+
+use warnings;
+use strict;
+
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
```diff
@@ -0,0 +1,117 @@
+=head1 LICENSE
+
+Copyright [2018-2024] EMBL-European Bioinformatics Institute
```
JAlvarezJarreta:
```suggestion
Copyright [2018-2025] EMBL-European Bioinformatics Institute
```

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
```diff
@@ -0,0 +1,117 @@
...
+use Moose;
+use Test::More;
+use Bio::EnsEMBL::DataCheck::Test::DataCheck;
+
+extends 'Bio::EnsEMBL::DataCheck::DbCheck';
+
+use constant {
+  NAME        => 'DisplayableSampleLocation',
+  DESCRIPTION => 'Sample location is displayable and has web_data attached to its analysis',
+  GROUPS      => ['analysis_description', 'core', 'brc4_core', 'geneset', 'meta_sample'],
```
JAlvarezJarreta:
```suggestion
  GROUPS      => ['analysis_description', 'core', 'geneset', 'meta_sample'],
```
Just a suggestion: the `brc4_core` group is no longer in use, but we have not removed it yet as we did not find it useful to do such a large patch for little gain.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
```diff
@@ -0,0 +1,117 @@
...
+        FROM meta
+       WHERE meta_key = 'genebuild.sample_location'
+         AND species_id = $species_id
+         AND meta_value NOT REGEXP '^.+:[0-9]+-[0-9]+\$'
+    /;
+
+  is_rows_zero($self->dba, $sql_2, $desc_2, $diag_2);
+
+  my $desc_3 = 'Sample location coordinates are properly ordered';
+  my $diag_3 = 'genebuild.sample_location start coordinate is greater than end coordinate';
```
JAlvarezJarreta, line 68: Since I'm not too familiar, worth asking: are we always choosing samples/coordinates from the "positive" strand?

JAlvarezJarreta (approved):
Makes sense to make, thank you for the new DC. Just a couple of comments, nothing major.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
JackCurragh: I _think_ that even if the target feature is on reverse strand the the coordinates need to be in this order <chr>:<smaller>-<bigger>. 

I basically just tried flipping them on one example location and the browser complained so I made it a DC.

lib/Bio/EnsEMBL/DataCheck/Checks/DisplayableSampleLocation.pm
JackCurragh: I just used another as a template for this set of DCs. Should have removed it myself so thank you!

leannehaggerty (changes requested):
The check looks good, but as it's a new check you will need to add it to index.json
https://github.com/Ensembl/ensembl-datacheck/blob/94cc3fb0f818d93c963c11639ef76cae00ce6c36/lib/Bio/EnsEMBL/DataCheck/index.json#L1355

leannehaggerty (approved):
Thank you!

PR #660, opened 2025-12-04
patch logic on number of rows expected from meta_key presence query

JackCurragh:
Correct broken logic where having 1 sample_location (desired number) led to failed datacheck.


manuelcarbajo (approved):
This seems fine: checking that the database contains exactly one genebuild.sample_location meta entry for this species, which is the intended test. Thanks

vinay-ebi (approved):

PR #661, opened 2025-12-04
patch logic of display xref check

JackCurragh:
Display xref check was looking for 0 rows from a `COUNT(*)` query to catch missing display xrefs when `COUNT(*)` will always return 1 row

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
@@ -53,11 +53,12 @@ sub tests {
   foreach my $type ("gene", "transcript") {
     my $desc = "${type}s have names set via display_xref_id";
     my $sql  = qq/
-      SELECT COUNT(*) FROM $type t
+      SELECT * FROM $type t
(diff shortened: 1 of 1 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
```diff
@@ -53,11 +53,12 @@ sub tests {
   foreach my $type ("gene", "transcript") {
     my $desc = "${type}s have names set via display_xref_id";
     my $sql  = qq/
-      SELECT COUNT(*) FROM $type t
+      SELECT * FROM $type t
```
vinay-ebi, line 56: Hi @JackCurragh, this data check looks good as I see no need to change sql query . 
test `is_rows_nonzero` is equivalent to `cmp_rows($dbc, $sql, '>', 0, $test_name)`  datacheck pass if row count is greater than zero .
This data check is advisory, so no impact on the handover service even if this fails.

JackCurragh: Yes, thank you @vinay-ebi for explaining.

PR #662, opened 2025-12-05
changed display xref exists from advisory to critical (for Vert genomes)

JackCurragh:
We expect all handed over cores from Genebuild pipelines to have gene symbols that come from the gene symbol transformer pipeline. 

I am opening this PR to propose moving this DC from `advisory` to `critical` but we need to ensure that the `skip_tests` is skipping what we wish it to skip. 

[Related Bug ticket](https://embl.atlassian.net/browse/EBD-73)

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
@@ -31,7 +31,7 @@ use constant {
   NAME           => 'DisplayXrefExists',
   DESCRIPTION    => 'At least one gene name exists',
   GROUPS         => ['core', 'xref', 'xref_gene_symbol_transformer', 'xref_name_projection'],
-  DATACHECK_TYPE => 'critical',
+  DATACHECK_TYPE => 'advisory',
   TABLES         => ['coord_system', 'gene', 'seq_region', 'transcript', 'xref'],
 };
 
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
@@ -31,7 +31,7 @@ use constant {
   NAME           => 'DisplayXrefExists',
   DESCRIPTION    => 'At least one gene name exists',
   GROUPS         => ['core', 'xref', 'xref_gene_symbol_transformer', 'xref_name_projection'],
-  DATACHECK_TYPE => 'advisory',
+  DATACHECK_TYPE => 'critical',
   TABLES         => ['coord_system', 'gene', 'seq_region', 'transcript', 'xref'],
 };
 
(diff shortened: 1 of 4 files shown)


JackCurragh:
One point of note is that the current proposed change will make transcript display xref ids mandatory. I don't believe we have any way to populate them currently..

leannehaggerty:
A couple of notes on this:

- EnsemblVertebrates will only cover the vertebrate databases, not the non-vertebrates that are generated via the genebuild pipelines and therefore have had GST run on them (team responsible might be a better filter here)
- Given that transcript names are not used anywhere, it might be best to split this check into 2:

1. transcript check that is advisory
2. gene check that is critical

- However, I also believe that the check should on whether a gene appears to have a name, i.e. something in the gene description (should confirm this) but does not have a display_xref_id, i.e. it's fine not to have display_xref_ids as long as there is nothing to display

Happy to discuss this further

JackCurragh:
Thanks Leanne. I have an update to this with it split in two in progress. I'll push those updates

The blocker until yesterday was that use of team responsible doesnt support checking the subset of Plantazoa genomes that have xrefs. Following discussion with @JAlvarezJarreta it was decided that their cores will only be subjected to this DC after they adopt a GST variant. So team responsible `Genebuild` will do for now. 

My understanding is that descriptions are populated at the same time as display xrefs using GST and as a result should always match. Which, now that I have written 'should', makes me realise you're dead right and this needs a DC. I will add it. 

While I am keeping the transcript check for now, I am actually against its inclusion on this branch. Perhaps it will be addressed in the move to python DCs but I think its important to have these cover everything in the core data model plus everything needed for the metadata db and not much more. 

That way DCs become an accurate record of what is currently important to Ensembl.

JackCurragh:
Sorry @leannehaggerty I was incorrect to say all would have `gene.display_xref_id` and `gene.description` populated at the same time. All RFAM originating annotations (tRNA, snoRNA, rRNA etc) will have descriptions but not display xrefs. 

If we filter to just gene biotype = protein coding, we find that mitochondial genes improted from RefSeq have descriptions but not xrefs. 

The best query I can come up with to reflect this is below. Let me know what you think but to me it seems quite rocky to make `critical` 
```
SELECT gene_id, display_xref_id, description
FROM gene g INNER JOIN seq_region sr USING (seq_region_id) 
INNER JOIN coord_system cs USING (coord_system_id) 
WHERE cs.species_id = 1   
AND g.description IS NOT NULL   
AND g.display_xref_id IS NULL 
AND description NOT LIKE '%RNA%' 
AND description NOT LIKE '%RFAM%' 
AND sr.name NOT LIKE 'MT';
```

JackCurragh:
I thinkw e need to get this reviewed and live rather than trying to get the xref description matching right as its not a simple problem.

dpopleton:
Have you tested this? Try a few with expected failures from genebuild, a few that will pass, a collection, and a metazoan/plant.
I am happy to improve and implement as the code looks good and the logic is reasonable.

JackCurragh:
@dpopleton Thanks to your prompting for validation and with Vinays help this DC is in much better shape. I have tested and confirmed that it will not run if genebuild.team_responsible is not defined (Although perhaps this needs to become a required DA). 

I have also validated that it will only run the test if team responsible is defined and the value is `Genebuild`. It will fail if all values in the `display_xref_id` column are `NULL`. 

This same logic is applied to every genome in every core.

vinay-ebi (approved):

dpopleton (approved):

PR #663, opened 2025-12-18
Fix/displayable location for multiple coord systems

JackCurragh:
Fix DisplayableSampleLocation false positive from multi-assembly seq_regions

The datacheck fails when multiple seq_regions share the same name across different assemblies (e.g., chromosome 4 in GRCm38, GRCm39, patches). The LEFT JOIN creates duplicate rows, and if any non-primary assembly lacks genes in the sample location coordinates, the IS NULL check incorrectly flags a failure.

Added coord_system join with cs.rank = 1 filter to restrict checks to primary assembly only.

Validated fix on mus_musculus_core_114_39 where three chromosome 4 seq_regions exist—only the primary (seq_region_id 125694) contains genes at the sample location coordinates.

### DIFF
--- a/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
+++ b/lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
@@ -31,7 +31,7 @@ use constant {
   NAME           => 'DisplayXrefExists',
   DESCRIPTION    => 'At least one gene name exists',
   GROUPS         => ['core', 'xref', 'xref_gene_symbol_transformer', 'xref_name_projection'],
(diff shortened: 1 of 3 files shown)


lib/Bio/EnsEMBL/DataCheck/Checks/DisplayXrefExists.pm
dpopleton, line 1: This should be okay, as it is only verts and you guys are responsible for it.
Have you tested this DC on a few of your cores (not just human, but some random other verts? Might want to try on a Plant or a microbe to make sure it autopasses)

dpopleton (approved):

